PR #3, opened 2020-03-21
Restructuring - 1 (Adding Data loader and associated Utility Functions).

kartikdutt18:
Basic Structure of models repo:

```
--models
        -- data (Empty everything will be download in runtime)

	 -- dataloaders (Input dataset name like "mnist" or path to your csv)

	 -- research (Train models from CLI)

		-- object_classification.cpp (specify model and params).

	-- models (All models implemented as a class)

		--lenet (All lenet models (1, 4, 5))
		--simple_nn (An FFN model)

	-- utils
		-- utils.hpp 
                -- Useful functions such as GetLabels to get labels for predictions
	-- weights
		-- Add weights for model (Download only, will be empty in repo)
	-- tests
		Added tests to download, load, train and perform inference using models.

```
There a ton to do, so I have added a list below of what I plan to accomplish in this PR:

- [x] Set basic structure ?
- [x] Add data loader for mnist.
- [x] Add Run Time Download of mnist using wget in cmake.
- [x] Add LeNet model.
- [x] Add Simple NN model. (I don't think this should be here, let me know what you think?)
- [x] Add Tests (BOOST_AUTO_TEST_CASE maybe ? , details in mlpack/examples#57)
- [ ] Add Documentation for models and how to use them.
- [x] Add object classification file. (This is where CLI would be added to specify parameters).
- [x] Add basic Cmake.
- [ ] Remove Cmake to match git submodules.
- [ ] Add downloadable weights (Also in run time).
- [x] Maybe Add LeNet versions as typedef ? (Sounds like a good idea, what do you think?)
- [x] According to discussion on IRC, I have removed LSTMs from here (I can add them as models if necessary, Let me know How I should proceed with this)
- [x] **Add comments and provide style fixes.** 
- [x] Change travis and maybe integrate Azure Pipelines?
I think there would be a separate PR to add CLI.

Adding a timeline here to complete remaining points to ensure that I don't fall behind:
- [x] Add object detection file (complete this by  27/03).
- [x] Add Typedef for lenet (by 27/03).
- [x] Train lenet models (by 27/03)
- [x] Add working tests (complete them by 28/03)
- [x] Document this and examples repo (complete by 29/03)
- [ ] Repeat for VAE (by 29/03).
This will complete the main C++ functionalities then.


shrit:
@kartikdutt18 it sounds a good idea of defining known architectures as `typedef` I even thought of defining a header file `LeNet.hpp` that defines the architecture which is separate from the `CLI` executables files. For example,  these models header files can be installed on the user machine. This will allow the user to do the following:

```cpp
#include <mlpack/models/LeNet.hpp>

int main()
{
  mlpack::models::LeNet le_net;
  le_net.Train(//callbacks can go here with data set as usual); // if you want to train
  le_net.Import_weights(/* Path for weights*/); // transfer learning
  le_net.another_functions()
}
```
Since these architectures are usually not modified, we can create a generic API for all these  models files.  What do you think @kartikdutt18 @rcurtin ?

kartikdutt18:
@shrit, That would be great. What I have imagined is that we can have a file called object_detection.cpp (which would have all CLI related stuff). And of course users would have access to models that are installed as you mentioned so that they can be used in some other files as well. So this way we can have ready to train and run models to solve classical problems as well as models that can be used for user specific problems as well.
I will be pushing some commits after write some unit tests.
Then we could do something like:
`./object_detection DATASET = path_to_dataset MODEL = model_name EPOCHS = 50 OTHER_PARAMS=...`

rcurtin:
Awesome work! :rocket: Looks good to me so far, just a few comments:

 - I think I agree that maybe a simple neural network isn't necessary here.  Someone could already use, e.g., mlpack's logistic regression implementation or random forest implementation if they're looking for a simple baseline.

 - I might suggest structuring the directories such that, e.g., there's `models/lenet5/` and this directory would contain *both* the `lenet5.hpp` file that defines the `LeNet5` class, and also the binding, e.g., `lenet5_main.cpp`, which defines the `mlpackMain()` and associated parameters that could be compiled into a `mlpack_lenet5` command-line program or `mlpack.lenet5()` Python function, etc.  The `research/` name doesn't make too much sense to me though.

Another thought might be, it could be good to do this as incremental one-at-a-time PRs instead of a single big one. :)  But up to you. :+1:

kartikdutt18:
I think that makes more sense to me, Maybe this makes sense : 

1. This PR to setup models, data loaders and tests.
2. Modify cmake for sub modules.
3. Add bindings to this repo.

I think that way we would have 3 PRs and each state we would have a working repo. @shrit, @rcurtin, What do you think?

kartikdutt18:
> The research/ name doesn't make too much sense to me though.

I couldn't think of any name so I borrowed the name from pytorch.

kartikdutt18:
Removed WIP tag, At least C++ portion is almost done. Just have to finish the tests and the error with linking boost tests(undefined main). If anyone has an idea what I might be missing, that would be great.

kartikdutt18:
Hey @rcurtin, @zoq, @shrit, Could you please review this PR or leave some general comments. I have got it working for the most part. I am training LeNet right now and I will add test for that and url for their download as well. As @zoq, I also plan on adding a download option through cmake as some people might want to download their datasets before runtime so that they directly run/test the model.
Few things that are left:
1. In some portion I have used pass-by-value (Change to pass by reference). (In progress)
2. I can run the tests using `./bin/models_test -t UtilsTest` however the Ctest is still empty(I have to figure that out).
3. Transfer VAE as well (majority done).
4. Add weights for model added. (In progress)

kartikdutt18:
The subsequent PRs would be about changing to git sub modules and then adding CLI.

dataloader/dataloader.hpp
```diff
@@ -0,0 +1,175 @@
...
+#include <dataloader/datasets.hpp>
+#include <mlpack/core.hpp>
+#include <utils/utils.hpp>
+
+
+#ifndef MODELS_DATALOADER_HPP
+#define MODELS_DATALOADER_HPP
+
+template<
+  typename DataSetX = arma::mat,
```
rcurtin, line 25: Might be easier to type as `DatasetX` (and lowercased `dataset` matches mlpack documentation's usage a bit better).  Your call. :+1:

dataloader/dataloader.hpp
```diff
@@ -0,0 +1,175 @@
...
+  DataSetY &ValidY() { return validY; }
+
+  //!Get the Scaler.
+  ScalerType Scaler() const { return scaler; }
+  //! Modify the Sclaer.
+  ScalerType &Scaler() { return scaler; }
+
+private:
+  // Utility Function to wrap indices.
+  size_t wrapIndex(int index, size_t length)
```
rcurtin, line 132:
```suggestion
  size_t WrapIndex(int index, size_t length)
```
(Just to match the mlpack style.  I didn't adapt usages of this function throughout the code though.)

dataloader/dataloader_impl.hpp
```diff
@@ -0,0 +1,142 @@
+/**
+ * @file dataloader_impl.hpp
+ * @author Eugene Freyman
+ * @author Mehul Kumar Nirala.
+ * @author Zoltan Somogyi
```
rcurtin, line 5:
```suggestion
```

I think this file is all you, correct me if I'm wrong. :)

kartikdutt18:
Also would it possible to add Azure pipelines and style-checks here as well. I will update travis-ci for the same.

dataloader/dataloader_impl.hpp
```diff
@@ -0,0 +1,142 @@
...
+>DataLoader<
+    DataSetX, DataSetY, ScalerType
+>::DataLoader(const std::string &dataset,
+              const bool shuffle,
+              const double ratio,
+              const bool useScaler,
+              const std::vector<std::string> augmentation,
+              const double augmentationProbability)
+{
+  if (dataset == "mnist")
```
rcurtin, line 47: Maybe it would be worthwhile to hold some struct that matches names of well-known datasets?  i.e.

```
// short name => details.  Maybe this can all be initialized in one line, I haven't tried.
const std::map<std::string, DatasetDetails> datasetMap;
datasetMap["mnist"] = Datasets::MNIST();
...
```

I think I have my syntax wrong there, but you get the idea I assume. :)

Then, in this function you could just check if `dataset` is in the keys of `datasets` and use that `DatasetDetails`, thus making the code below not specific to MNIST. :+1:

dataloader/dataloader_impl.hpp
```diff
@@ -0,0 +1,142 @@
...
+    {
+      Utils::DownloadFile(Datasets::MNIST().testDownloadUrl, "./../data/mnist_test.csv",
+          "mnist_test.csv");
+
+      if (!Utils::CompareSHA256("./../data/mnist_test.csv", Datasets::MNIST().testHash))
+        std::cout << "Corrupted Test Data Downloaded." << std::endl;
+    }
+
+    LoadCSV("./../data/mnist_test.csv", false, false, useScaler, true, 0, -1);
+  }
```
rcurtin, line 73: Maybe is it worth adding this here so the function doesn't ever return without doing anything?
```
else
{
  Log::Fatal << "Unknown dataset '" << dataset << "'!" << std::endl;
}
```

dataloader/dataloader_impl.hpp
```diff
@@ -0,0 +1,142 @@
...
+  else
+  {
+    if (useScaler)
+    {
+      //scaler.Transform(dataset, dataset);
+    }
+
+    testX = dataset.submat(wrapIndex(startInputFeatures, dataset.n_rows),
+      0, wrapIndex(endInputFeatures, dataset.n_rows), dataset.n_cols - 1);
```
rcurtin, line 137:
```suggestion
        0, wrapIndex(endInputFeatures, dataset.n_rows), dataset.n_cols - 1);
```

dataloader/dataloader_impl.hpp
```diff
@@ -0,0 +1,142 @@
...
+        wrapIndex(endPredictionFeatures, validDataset.n_rows));
+
+    // Add support for augmentation here.
+    std::cout << "Training Dataset Loaded." << std::endl;
+  }
+  else
+  {
+    if (useScaler)
+    {
+      //scaler.Transform(dataset, dataset);
```
rcurtin, line 133: Any particular reason this is commented out?  It seems like you would want to fit the scaler to the full dataset here.

dataloader/datasets.hpp
```diff
@@ -0,0 +1,58 @@
...
+class Datasets
+{
+  public:
+  const static DatasetDetails MNIST()
+  {
+    return DatasetDetails("mnist",
+      "https://raw.githubusercontent.com/kartikdutt18/mlpack-models-weights-and-datasets/master/mnist-dataset/mnist_train.csv",
+      "https://raw.githubusercontent.com/kartikdutt18/mlpack-models-weights-and-datasets/master/mnist-dataset/mnist_test.csv",
+      "cc10cd2dcac4fa2b67b8e9b7c90019cb8669668ed11a1ea71d980418785b5b11",
+      "1e4c6240156c2316c012a655ff268e8b7f37b4a4dabe49853ab3a60af0ed1bca");
```
rcurtin, line 52: We can make this use an mlpack.org URL before merge. :+1:

examples/object_classification/object_classification.cpp
```diff
@@ -0,0 +1,45 @@
...
+#include <models/lenet/lenet.hpp>
+#include <ensmallen.hpp>
+
+using namespace mlpack;
+using namespace mlpack::ann;
+using namespace arma;
+using namespace std;
+using namespace ens;
+
+int main()
```
rcurtin, line 18: I guess, are you thinking that we can add the `PARAM_*()` calls in a later PR?

examples/object_classification/object_classification.cpp
```diff
@@ -0,0 +1,45 @@
...
+  model.Train(dataloader.TrainX(),
+              dataloader.TrainY(),
+              optimizer,
+              ens::PrintLoss(),
+              ens::ProgressBar(),
+              ens::EarlyStopAtMinLoss());
+
+  module1.SaveModel("./../weights/lenet.bin");
+  module1.LoadModel("./../weights/lenet.bin");
+}
```
rcurtin, line 45: Can you make sure the files end with a newline?  (Sorry, minor issue. :))

examples/object_classification/object_classification.cpp
```diff
@@ -0,0 +1,45 @@
+#include <mlpack/core.hpp>
```
rcurtin, line 1: I think it would be useful to add some comments to the top of this file describing what it's for, the type of model that it trains, how one might use it, etc.

models/lenet/lenet.hpp
```diff
@@ -0,0 +1,202 @@
...
+   * @param inputHeight Height of the input image.
+   * @param numClasses Optional number of classes to classify images into,
+   *                   only to be specified if includeTop is  true.
+   * @param weights One of 'none', 'imagenet'(pre-training on mnist) or path to weights.
+   */
+  LeNet(const size_t inputChannel,
+        const size_t inputWidth,
+        const size_t inputHeight,
+        const size_t numClasses = 1000,
+        const std::string &weights = "none");
```
rcurtin, line 66:
```suggestion
        const std::string& weights = "none");
```
Tiny style comment. :)

models/lenet/lenet.hpp
```diff
@@ -0,0 +1,202 @@
...
+   */
+  size_t PoolOutSize(const size_t size,
+                     const size_t k,
+                     const size_t s)
+  {
+    return std::floor(size - 1) / s + 1;
+  }
+
+  //! Locally stored LeNet Model.
+  Sequential<> *leNet;
```
rcurtin, line 179:
```suggestion
  Sequential<>* leNet;
```
Also, maybe it would be better to store `leNet` as an `FFN` instead?  (If one wanted to use `LeNet` inside of other models, it would be easy to wrap it in a `Sequential<>` then.  But at least `object_recognition.cpp` uses an FFN that's only `LeNet`, so that code could be simplified with `leNet` as just an `FFN`. :+1:

models/lenet/lenet.hpp
```diff
@@ -0,0 +1,202 @@
...
+#include <mlpack/methods/ann/layer/layer_types.hpp>
+#include <mlpack/methods/ann/init_rules/random_init.hpp>
+
+namespace mlpack {
+namespace ann /** Artificial Neural Network. */{
+
+/**
+ * Definition of a LeNet CNN.
+ * 
+ * @param leNetVer Version of LeNet.
```
rcurtin, line 43: I think this can only be 1, 4, or 5.  Want to add a static assert check for that?

models/lenet/lenet_impl.hpp
```diff
@@ -0,0 +1,105 @@
...
+                       inputWidth(std::get<1>(inputShape)),
+                       inputHeight(std::get<2>(inputShape)),
+                       numClasses(numClasses),
+                       weights(weights)
+{
+  leNet = new Sequential<>();
+  ConvolutionBlock(inputChannel, 6, 5, 5, 1, 1, 2, 2);
+  PoolingBlock(2, 2, 2, 2);
+  ConvolutionBlock(6, 16, 5, 5, 1, 1, 2, 2);
+  PoolingBlock(2, 2, 2, 2);
```
rcurtin, line 53: I'm confused, these lines don't seem to do anything?

models/lenet/lenet_impl.hpp
```diff
@@ -0,0 +1,105 @@
...
+    leNet->Add<Linear<>>(120, numClasses);
+  }
+  else if (leNetVer == 5)
+  {
+    leNet->Add<Linear<>>(16 * inputWidth * inputHeight, 120);
+    leNet->Add<LeakyReLU<>>();
+    leNet->Add<Linear<>>(120, 84);
+    leNet->Add<LeakyReLU<>>();
+    leNet->Add<Linear<>>(84, 10);
+  }
```
rcurtin, line 72: If `leNetVer` is a template parameter, it might make the most sense to simply split these out into separate functions, like:

```
template<>
LeNet<1>::BuildNetwork()
{
  leNet->Add<Linear<>>(16 * inputWidth * inputHeight, numClasses);
}

template<>
LeNet<4>::BuildNetwork()
{
  leNet->Add<Linear<>>(16 * inputWidth * inputHeight, 120);
  leNet->Add<LeakyReLU<>>();
  leNet->Add<Linear<>>(120, numClasses);
}

...
```

Then you can just call `BuildNetwork()` from this constructor and it will select the right specialization. :)

models/lenet/lenet.hpp
```diff
@@ -0,0 +1,202 @@
...
+
+  /**
+   * LeNet constructor intializes input shape and number of classes.
+   *
+   * @param inputChannels Number of input channels of the input image.
+   * @param inputWidth Width of the input image.
+   * @param inputHeight Height of the input image.
+   * @param numClasses Optional number of classes to classify images into,
+   *                   only to be specified if includeTop is  true.
+   * @param weights One of 'none', 'imagenet'(pre-training on mnist) or path to weights.
```
rcurtin, line 60: What are the requirements for `weights` here?  I see from the code that if it's `mnist`, we'll load MNIST weights from a file, but what does a user have to run to make sure that the files exist on disk?  It's probably worth pointing that out here in the documentation.  (I don't see any code handling for `imagenet` though?)

tests/dataloader_tests.cpp
```diff
@@ -0,0 +1,33 @@
...
+ */
+BOOST_AUTO_TEST_CASE(MNISTDataLoaderTest)
+{
+  DataLoader<> dataloader("mnist", true, 0.80);
+  // Check for correct dimensions.
+  BOOST_REQUIRE_EQUAL(dataloader.TrainX().n_cols, 784);
+  BOOST_REQUIRE_EQUAL(dataloader.TestX().n_cols, 784);
+  BOOST_REQUIRE_EQUAL(dataloader.ValidX().n_cols, 784);
+  BOOST_REQUIRE_EQUAL(dataloader.TrainX().n_rows, 33600);
+}
```
rcurtin, line 31: Want to add some tests for when `DataLoader` is used incorrectly?  (i.e., does it crash as expected?)

tests/utils_tests.cpp
```diff
@@ -0,0 +1,51 @@
+/**
+ * @file dataloader_tests.cpp
```
rcurtin, line 2:
```suggestion
 * @file utils_tests.cpp
```
:)

tests/utils_tests.cpp
```diff
@@ -0,0 +1,51 @@
...
+ */
+BOOST_AUTO_TEST_CASE(DownloadFileTest)
+{
+  Utils::DownloadFile(Datasets::MNIST().trainDownloadUrl, "./../data/mnist_train.csv", "", false);
+  BOOST_REQUIRE(Utils::PathExists("./../data/mnist_train.csv"));
+  BOOST_REQUIRE(Utils::CompareSHA256("./../data/mninst_train.csv", Datasets::MNIST().trainHash));
+
+  Utils::DownloadFile(Datasets::MNIST().testDownloadUrl, "./../data/mnist_test.csv", "", false);
+  BOOST_REQUIRE(Utils::PathExists("./../data/mnist_test.csv"));
+  BOOST_REQUIRE(Utils::CompareSHA256("./../data/mninst_test.csv", Datasets::MNIST().testHash));
```
rcurtin, line 30: Should we remove the files we downloaded with `remove()` after running the tests?  (Also should this line compare with `mnist_test.csv` not `mninst_test.csv`?)

utils/utils.hpp
```diff
@@ -0,0 +1,73 @@
...
+      std::cout << "Downloading "<< name << std::endl;
+    }
+
+    std:: string command = "curl " + url + " -o " + fileName;
+    if (progressBar)
+      command += " --progress-bar";
+    else
+      command += " --silent";
+    
+    return system((const char *)command.c_str());
```
rcurtin, line 59: Ah, I think this will only work on Linux or OS X (or systems where `curl` is on the path, which is probably not usually the case on Windows).  Do you think you can use the cURL C API for this, then link against libcurl?

utils/utils.hpp
```diff
@@ -0,0 +1,73 @@
...
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ */
+
+#ifndef MODELS_UTILS_HPP
+#define MODELS_UTILS_HPP
+
+#include <iostream>
+#include <cstdlib>
+#include <sys/stat.h>
+#include <openssl/sha.h>
```
rcurtin, line 19: If we need OpenSSL, I think that's just fine, but should we add it to the dependencies we search for and require in the main `CMakeLists.txt`?

dataloader/CMakeLists.txt
```diff
@@ -0,0 +1,27 @@
...
+    dataloader.hpp
+    dataloader_impl.hpp
+)
+
+if(DEBUG)
+  message("Compilation with debug info (with ggdb3 flag)")
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -ggdb3")
+else()
+  message("Compilation without debug info (without ggdb3 flag)")
+endif()
```
rcurtin, line 19: Seems like this should be added to the root CMakeLists.txt?

zoq:
@mlpack-jenkins test this please

zoq:
@mlpack-jenkins test this please

zoq:
About setting up azure pipelines, we would need a build strategy first.

zoq:
Also, once https://github.com/mlpack/jenkins-conf/pull/18 is merged, the style check should work correctly.

kartikdutt18:
Great, Also Can I also get your opinion on this PR. It would really mean a lot. Thanks.

kartikdutt18:
> About setting up azure pipelines, we would need a build strategy first.

Does this refer to yaml for running azure pipelines ? I can add that as well here but that will definitely be changed when this repo switches to submodules. I am probably missing something here.

zoq:
Yes, once the yaml file there, I can point azure to it.

kartikdutt18:
> Yes, once the yaml file there, I can point azure to it.

Great, I will add in the next couple of days.

zoq:
@mlpack-jenkins test this please

kartikdutt18:
Sorry I got caught up in codejam and an assignment submission that was due today. Should I resolve conflicts maybe that is causing the issue?

zoq:
@mlpack-jenkins test this please

zoq:
Shouldn't be a problem

> Sorry I got caught up in codejam and an assignment submission that was due today. Should I resolve conflicts maybe that is causing the issue?

No need, we don't build the code, looks like it's fixed now.

kartikdutt18:
Ahh great, Thanks a lot. Will fix make all changes tonight.

utils/utils.hpp
kartikdutt18: I am going a different way at least for now, I am writing my own implementation of checksum.

utils/utils.hpp
kartikdutt18: Done.

tests/utils_tests.cpp
kartikdutt18: Sounds like a good idea. Will add it to the implementation.

tests/dataloader_tests.cpp
kartikdutt18: Agreed, I will add them as well.

kartikdutt18:
Hey @rcurtin, I think we discussed most of the points in the video meet, However here is the summary incase I missed something.

> What's the use case for object_recognition.cpp?

The current file will be adapted in another PR to support CLI. To specify model, training params and dataset from CLI so that a user can train a model from command line on any dataset.
Added a comment on top of the file to mention the same.

> leNet as just an FFN.

I am testing the implementation as FFN right now, once it works I will push it as well.

> Having a build network function.

I haven't implemented that as it will introduce redundant code as various versions of the model have the same building blocks. Incase I have missed something, kindly let me know.

I have also read up a bit about azure pipelines and I will add a commit for build tomorrow as well.

I have a long weekend this week, I plan on completing this PR during the same.

I will take care of style issues together (I hope that's okay). Sorry.
I have also replied to some comments that I haven't resolved yet. Thanks a lot for the review.

models/lenet/lenet_impl.hpp
kartikdutt18: I did something similar to how PyTorch builds model. Convolution and Pooling Blocks add convolutional layer, activation function layer pooling layer and normalization respectively. Rather than writing them again and again we can simply use blocks to remove redundant code. They add layers to the leNet layer.

dataloader/dataloader.hpp
```diff
@@ -0,0 +1,215 @@
...
+   * Constructor for DataLoader. This is used for loading popular Datasets such as
+   * MNIST, ImageNet, Pascal VOC and many more.
+   * 
+   * @param datasetPath Path or name of dataset.
+   * @param shuffle whether or not to shuffle the data.
+   * @param ratio Ratio for train-test split.
+   * @param augmentation Adds augmentation to training data only.
+   * @param augmentationProbability Probability of applying augmentation on dataset.
+   */
+  DataLoader(const std::string &dataset,
```
zoq, line 45: Looks like the comments don't match with the parameters, but maybe this isn't ready yet?

kartikdutt18: > Looks like the comments don't match with the parameters

Terribly sorry about, the dataloader went through different versions to remove redundant / duplicate code and I think that led to incomplete parameters in it.

> but maybe this isn't ready yet?

There are only few points left. I was writing more tests for models, utils and dataloader. In the next few hours I will remove all dependencies and switch to boost lib. I will review my PR and make the required style changes like this one. Maybe you could take a look after that if you get a chance. Thanks a lot.

kartikdutt18:
Hey @rcurtin, @zoq, Would there be any copyright issues if use a portion of code from boost examples from the following:  [link 1](https://www.boost.org/doc/libs/1_55_0/doc/html/boost_asio/example/cpp03/http/client/async_client.cpp) and [link 2](https://www.boost.org/doc/libs/1_55_0/doc/html/boost_asio/example/cpp03/http/client/async_client.cpp).

zoq:
> Hey @rcurtin, @zoq, Would there be any copyright issues if use a portion of code from boost examples from the following: [link 1](https://www.boost.org/doc/libs/1_55_0/doc/html/boost_asio/example/cpp03/http/client/async_client.cpp) and [link 2](https://www.boost.org/doc/libs/1_55_0/doc/html/boost_asio/example/cpp03/http/client/async_client.cpp).

That is a pretty minimal example, so I don't think we will run into any problem here, but we might want to mention the examples (add a link).

kartikdutt18:
> That is a pretty minimal example, so I don't think we will run into any problem here, but we might want to mention the examples (add a link).

Ahh, great. I will update them with style fixes in the next commit.

kartikdutt18:
Hey @zoq, @rcurtin, If you get a chance, could you please take another look at this. I have fixed all style issues pertaining to the files that I have added. Sorry for the delayed response, I had to prepare for my online assessment.  I'll try to finish up as much as I can during the same period. I will push the yaml files as well. I would really appreciate your suggestions. Thanks a ton.

kartikdutt18:
For style changes in this repo and examples that are referring to the current files, I can open another PR.

dataloader/dataloader.hpp
```diff
@@ -0,0 +1,225 @@
...
+#include <mlpack/core.hpp>
+#include <utils/utils.hpp>
+
+
+#ifndef MODELS_DATALOADER_HPP
+#define MODELS_DATALOADER_HPP
+
+/**
+ * Dataloader class to load popular datasets.
+ *
```
zoq, line 26: Maybe a simple example here would be nice to have, similar to:

https://github.com/mlpack/mlpack/blob/9672d5d15368aca5600e6f72cb8f2fd15d2bc25d/src/mlpack/methods/mean_shift/mean_shift.hpp#L32-L41

dataloader/dataloader.hpp
```diff
@@ -0,0 +1,225 @@
...
+   * MNIST, ImageNet, Pascal VOC and many more.
+   * 
+   * @param datasetPath Path or name of dataset.
+   * @param shuffle whether or not to shuffle the data.
+   * @param ratio Ratio for train-test split.
+   * @param useScaler Use feature scaler for pre-processing the dataset.
+   * @param augmentation Adds augmentation to training data only.
+   * @param augmentationProbability Probability of applying augmentation on dataset.
+   */
+  DataLoader(const std::string& dataset,
```
zoq, line 53: Do you think we should provide a Constructor that returns the loaded and processed dataset as well?

dataloader/dataloader_impl.hpp
```diff
@@ -0,0 +1,151 @@
...
+    if (datasetMap[dataset].loadCSV)
+    {
+      LoadCSV(datasetMap[dataset].trainPath, true, shuffle, ratio, useScaler,
+              datasetMap[dataset].dropHeader,
+              datasetMap[dataset].startTrainingInputFeatures,
+              datasetMap[dataset].endTrainingInputFeatures,
+              datasetMap[dataset].endTrainingPredictionFeatures,
+              datasetMap[dataset].endTrainingPredictionFeatures);
+
+      if (dataset == "mnist")
```
zoq, line 59: We might want to apply a policy based design here, like a dataset specific class to apply some specific functions after the data is loaded, something like:

```c++
class Mnist {
  static void Update(...)
  {

  }
}
```

let me know what you think

dataloader/dataloader.hpp
kartikdutt18: Yes, will add it in the next commit. Thanks a lot for the review.

dataloader/dataloader_impl.hpp
kartikdutt18: Yeah, we could do that. Just one more thing, I am currently storing a class that determines type of dataset and if it is of `csv` format we store rows to drop, rows to use for prediction etc. I think we should keep that as it is or would you suggest having them in the class you mentioned above.

kartikdutt18: This would definitely more sense when more datasets are added.

dataloader/dataloader.hpp
kartikdutt18: I am terribly sorry but I didn't get your point here. I assuming it means that we should add a constructor for custom dataset as well ?

dataloader/dataloader_impl.hpp
kartikdutt18: Hey @zoq, I was thinking if we could do something like the following :

```
class Preprocess
{
   public:
   static void MNIST(input-args)
   {
     // Preprocess dataset.
    }
}
```

That way we don't have to have a separate class for each dataset we add here, and eventually save the lines that would have to be maintained.

Kindly let me know if this makes sense. Thanks a lot.

Regards,
Kartik.

dataloader/dataloader.hpp
zoq: Nevermind I thought I have to first create the `DataLoader` object and after that use `DownloadDataset` to get the dataset, but I can just use `TrainX`. Sorry for the confusion.

dataloader/dataloader_impl.hpp
zoq: Sure, that looks reasonable to me.

kartikdutt18: Great, Thanks a lot.

dataloader/dataloader.hpp
kartikdutt18: Great, making rest of the changes now. Thanks a ton!

kartikdutt18:
Hey @zoq, when you get a chance could point the azure pipelines to the yaml file here. I have only added OSX for now and I'll add Windows and Linux in the subsequent commits.
Thanks,
Regards.

zoq:
Great, we just need an initial config file, so once https://github.com/mlpack/models/pull/7 is merged I can set up the pipeline.

kartikdutt18:
Great, Thanks a lot.

zoq:
Looks like there are some errors in the ci.yaml file but the pipeline works now.

dataloader/dataloader.hpp
```diff
@@ -0,0 +1,245 @@
...
+               const int startInputFeatures = -1,
+               const int endInputFeatures = -1,
+               const int startPredictionFeatures = -1,
+               const int endPredictionFeatures = -1,
+               const std::vector<std::string> augmentation =
+                   std::vector<std::string>(),
+               const double augmentationProbability = 0.2);
+
+  //! Get the Training Dataset.
+  DatasetX TrainX() const { return trainX; }
```
zoq, line 122: I'm wondering if we should use a more descriptive name, like `TrainSet` or `TrainPredictors` and `TrainLabels` or `TrainResponses`, let me know what you think.

zoq: I guess we could also return a tuple.

kartikdutt18: Yes, Maybe `TrainFeatures` ? and `TrainLabels` make sense.

> I guess we could also return a tuple.

Yes, that could be added.
Thanks for the review.
Regards.

zoq: I like the naming.

dataloader/dataloader.hpp
```diff
@@ -0,0 +1,245 @@
...
+  //! Get the Scaler.
+  ScalerType Scaler() const { return scaler; }
+  //! Modify the Sclaer.
+  ScalerType& Scaler() { return scaler; }
+
+ private:
+  /**
+   * Downloads and checks hash for given dataset.
+   */
+  void DownloadDataset(const std::string& dataset)
```
zoq, line 161: Might be nice to comment on the parameter even if it's a private function.

dataloader/dataloader.hpp
```diff
@@ -0,0 +1,245 @@
...
+    }
+    if (!Utils::PathExists(datasetMap[dataset].testPath))
+    {
+      Utils::DownloadFile(datasetMap[dataset].trainDownloadUrl,
+          datasetMap[dataset].testPath, dataset + "_testing_data.",
+          false);
+
+      if (!Utils::CompareCRC32(datasetMap[dataset].testPath,
+          datasetMap[dataset].testHash))
+        mlpack::Log::Fatal << "Corrupted Testing Data for " <<
```
zoq, line 184: I would use `{}` here to make this easier to read.

dataloader/dataloader.hpp
```diff
@@ -0,0 +1,245 @@
...
+            dataset << " Downloaded." << std::endl;
```
zoq, line 185: Should this be `downloaded.`?

dataloader/dataloader_impl.hpp
```diff
@@ -0,0 +1,147 @@
...
+  arma::mat dataset;
+  data::Load(datasetPath, dataset, true);
+
+  dataset = dataset.submat(0, size_t(dropHeader), dataset.n_rows - 1,
+      dataset.n_cols - 1);
+
+  if (loadTrainData)
+  {
+    arma::mat trainDataset, validDataset;
+    data::Split(dataset, trainDataset, validDataset, ratio /* Add shuffle*/);
```
zoq, line 106:
```suggestion
    data::Split(dataset, trainDataset, validDataset, ratio /* Add shuffle */);
```

dataloader/dataloader_impl.hpp
```diff
@@ -0,0 +1,147 @@
...
+
+    validX = validDataset.rows(WrapIndex(startInputFeatures,
+        validDataset.n_rows), WrapIndex(endInputFeatures,
+        validDataset.n_rows));
+
+    validY = trainDataset.rows(WrapIndex(startPredictionFeatures,
+        validDataset.n_rows), WrapIndex(endPredictionFeatures,
+        validDataset.n_rows));
+
+    // Add support for augmentation here.
```
zoq, line 131: We might want to add a TODO here, to make this easier to find.

dataloader/dataloader_impl.hpp
```diff
@@ -0,0 +1,147 @@
...
+  }
+  else
+  {
+    if (useScaler)
+    {
+      scaler.Transform(dataset, dataset);
+    }
+
+    testX = dataset.submat(WrapIndex(startInputFeatures, dataset.n_rows),
+      0, WrapIndex(endInputFeatures, dataset.n_rows), dataset.n_cols - 1);
```
zoq, line 142:
```suggestion
        0, WrapIndex(endInputFeatures, dataset.n_rows), dataset.n_cols - 1);
```

dataloader/dataloader_impl.hpp
```diff
@@ -0,0 +1,147 @@
...
+    std::cout << "Testing Dataset Loaded." << std::endl;
```
zoq, line 143: We could also use `mlpack::log` here, but I don't really have a preference.

dataloader/preprocessor.hpp
```diff
@@ -0,0 +1,39 @@
...
+ * @tparam DatasetY Datatype for prediction features.
+ */
+template<
+  typename DatasetX = arma::mat,
+  typename DatasetY = arma::mat
+>
+class PreProcessor
+{
+ public:
+  static void MNIST(DatasetX& /*trainX*/, DatasetY& trainY,
```
zoq, line 31:
```suggestion
  static void MNIST(DatasetX& /* trainX */, DatasetY& trainY,
```

Just to be consistent with the rest.

dataloader/preprocessor.hpp
```diff
@@ -0,0 +1,39 @@
...
+      DatasetY& /*validX*/, DatasetY& validY, DatasetX& /*testX*/)
```
zoq, line 32:
```suggestion
      DatasetY& /* validX */, DatasetY& validY, DatasetX& /* testX */)
```

models/lenet/lenet.hpp
```diff
@@ -0,0 +1,210 @@
+/**
+ * @file lenet.hpp
+ * @author Eugene Freyman
+ * @author Daivik Nema
+ * @author Kartik Dutt
+ * 
+ * Definition of LeNet generally used for object detection.
```
zoq, line 7: I'm wondering if we should split this up into multiple PR's makes it easier to review, let me know what you think.

dataloader/dataloader.hpp
kartikdutt18: Sure, Sorry I missed it.

dataloader/dataloader.hpp
kartikdutt18: Agreed.

dataloader/dataloader.hpp
kartikdutt18: Yes. Thanks for pointing it out.

dataloader/dataloader_impl.hpp
kartikdutt18: Will do.

dataloader/dataloader.hpp
```diff
@@ -0,0 +1,245 @@
...
+
+  //! Get the Training Dataset.
+  DatasetX TrainX() const { return trainX; }
+
+  //! Modify the Training Dataset.
+  DatasetX& TrainX() { return trainX; }
+
+  //! Get the Training Dataset.
+  DatasetY TrainY() const { return trainY; }
+  //! Modify the Training Dataset.
```
saksham189, line 129: I think the documentation should be `Modify the Training Dataset Labels.`
and `Modify the Training Dataset Predictors.` 

Could you make the changes everywhere else?

dataloader/dataloader.hpp
```diff
@@ -0,0 +1,245 @@
...
+  DatasetX& ValidX() { return validX; }
+
+  //! Get the Validation Dataset.
+  DatasetY ValidY() const { return validY; }
+  //! Modify the Validation Dataset.
+  DatasetY& ValidY() { return validY; }
+
+  //! Get the Scaler.
+  ScalerType Scaler() const { return scaler; }
+  //! Modify the Sclaer.
```
saksham189, line 154: Scaler

dataloader/dataloader.hpp
```diff
@@ -0,0 +1,245 @@
...
+  //! Locally stored path of dataset.
+  std::string testDatasetPath;
+
+  //! Locally stored ratio for train-test split.
+  double ratio;
+
+  //! Locally stored augmentation.
+  std::vector<std::string> augmentation;
+
+  //! Locally stored augmented probability.
```
saksham189, line 239: IMO `augmentation probability` would sound better.

models/lenet/lenet.hpp
saksham189: Yes I agree with @zoq This PR is already quite big. 
Let us know what you think @kartikdutt18

kartikdutt18: Agreed, as discussed on the IRC, I will remove the portion related to LeNet and object detection and keep only the data loader portion with its utility functions. I hope that makes sense. Let me know what you think.

dataloader/dataloader.hpp
kartikdutt18: Sorry about that, will do.

dataloader/dataloader.hpp
kartikdutt18: Agreed, Will make the changes. Thanks a lot.

dataloader/dataloader.hpp
kartikdutt18: Sorry about that, will make the required changes. Thanks a lot.

kartikdutt18:
Great. Thanks, I'll shorten the PR shortly. Thanks a lot.

kartikdutt18:
Hey @zoq, @saksham189, It seems like I am unable to create / write file (in runtime) testing in azure pipelines. I added a bunch of warning during download to see if it created files as well as wrote a test which gives a warning if the machine was unable to create a new file. [Both](https://dev.azure.com/mlpack/mlpack/_build/results?buildId=1816&view=logs&j=1b38ea94-fede-5667-20eb-44d3de1ec085&t=40b5b531-4362-543b-1da6-f0cdf2a5edbe&l=26) of these warnings appear in CTest.
This works fine locally, I am most likely missing something, Could you please help me out with this.
Thanks a ton.

tests/dataloader_tests.cpp
```diff
@@ -0,0 +1,68 @@
...
+/**
+ * Simple test for Dataloader.
+ */
+BOOST_AUTO_TEST_CASE(CSVDataLoaderTest)
+{
+  // Check that dataloader loads only known datasets.
+  BOOST_REQUIRE_THROW(DataLoader<>("no-dataset", true), std::runtime_error);
+
+  // Check Load CSV Function for Dataloader.
+  Utils::DownloadFile("/datasets/iris.csv", "./../data/iris.csv");
```
zoq, line 30: Can you check if that works with an absolute path? Or I guess `iris.csv` should work as well, maybe the path isn't correct? Looking at the ci config:

`script: cd build/tests/ && sudo CTEST_OUTPUT_ON_FAILURE=1 ctest -R UtilsTest`

we should be in `build/tests/` so I would think we have to use `"./../../data/iris.csv"`?

.ci/ci.yaml
```diff
@@ -4,12 +4,41 @@ trigger:
...
-  vmImage: 'ubuntu-latest'
+  steps:
+  - template: linux-steps.yaml
 
-steps:
-- script: echo Hello, checkout mlpack/models!
-  displayName: 'Initial ci config.'
+# Windows build.
+# I'll add this once I understand a bit more about powershell commands.
+# Not a windows user. Sorry.
```
zoq, line 44: I think that is true for most of us 😃

tests/dataloader_tests.cpp
kartikdutt18: > we should be in build/tests/ so I would think we have to use "./../../data/iris.csv"?

I think this shouldn't be required since CtestConfig.CMake sets path for execution to build so the data folder should be accessible. I'll try with absolute path and get back to you. Thanks a lot.

.ci/ci.yaml
kartikdutt18: True, xD

tests/dataloader_tests.cpp
kartikdutt18: Finally figured it out, The tests pass now (getting 400 in one of them) but the rest pass. The issue was GitHub deleted the empty data folder. Locally I still had the folder so that caused this issue on azure and not locally. Sorry about this.
Thanks a lot.
Regards.

zoq: Ah, great that you figured it out.

kartikdutt18: I made a lot of commits for this. Hopefully this commit will give a green build on OSX. (Squashing the rest).

zoq: Right, it's often not that easy to debug something on the ci pipeline.

zoq:
So, I think with the build issues solved, what is left is to have some form of documentation, or do you like to add that in another PR?

kartikdutt18:
> So, I think with the build issues solved, what is left is to have some form of documentation, or do you like to add that in another PR?

I could do that in another PR, I tried reviewing this and it's a bit too long. I also have the image dataloader, lenet PR almost ready so once this done I can open them sequentially. I also need to add windows build before we merge this. Let me know what you think.

zoq:
Sounds good, let's integrate the windows ci config here and open other PR's for each extension. Nice work!

zoq:
Let me know if you need any help with the windows build.

kartikdutt18:
Hey @zoq, I did try a few things today so I was able to configure armadillo in models repo and mlpack. I am still facing issues with linking mlpack to models repo which hopefully I'll be able to resolve by tomorrow, if you have any suggestions to fix the build that would be great. Thanks a lot for your time.

kartikdutt18:
Hey @zoq, The windows is built is almost done, The build links everything that is needed However it fails because of the following [error](https://dev.azure.com/mlpack/mlpack/_build/results?buildId=1870&view=logs&j=9bb75cf0-500a-52c4-dbba-9c305edd1534&t=8b7b5019-6a91-5c78-a202-396e308acfa9&l=40) :

```
file STRINGS file
  "D:/a/1/s/mlpack/build/Release/include/mlpack/core/util/version.hpp" cannot
  be read.
```

Once this is resolved the windows build should pass as well. Could you point me to the direction in which I should look to fix the above error.

kartikdutt18:
Ahh, so the windows build is fixed now.

.ci/windows-steps.yaml
```diff
@@ -0,0 +1,124 @@
...
+    -DMLPACK_INCLUDE_DIR="..\mlpack\build\Release\include" `
+    -DMLPACK_LIBRARY="..\mlpack\build\Release\lib" `
+    -DBOOST_INCLUDEDIR=$(Agent.ToolsDirectory)\boost.1.60.0.0\lib\native\include `
+    -DBOOST_LIBRARYDIR=$(Agent.ToolsDirectory)\boost_libs ..
+  displayName: 'Build models'
+
+# Run tests via ctest.
+- bash: |
+    cd build/tests
+    CTEST_OUTPUT_ON_FAILURE=1 ctest -T Test -C Release .
```
kartikdutt18, line 115:
```suggestion
    CTEST_OUTPUT_ON_FAILURE=1 ctest -T Test -C ./. 
```

kartikdutt18:
I have opened  mlpack/mlpack#2409 to discuss windows build.
Regards.

.ci/windows-steps.yaml
```diff
@@ -0,0 +1,141 @@
...
+  displayName: 'Configure models'
+
+# Build models
+- task: MSBuild@1
+  inputs:
+    solution: 'build/*.sln'
+    msbuildLocationMethod: 'location'
+    msbuildVersion: $(MSBuildVersion)
+    configuration: 'Release'
+    msbuildArchitecture: 'x64'
```
zoq, line 123: Let's switch to `x86` and see if that helps.

Yashwants19: One more thing @zoq we are also getting warning regarding MLPACK_LIBRARY while configuration. Should we change that too? 
I think so It must be `mlpack/build/Release/mlpack.lib`.

zoq: Yes, good idea.

kartikdutt18: Sure, I'll make the change. Thanks a lot.

Yashwants19: Hey @kartikdutt18 MLPACK_INCLUDE_DIR was looking good. You just have to change MLPACK_LIBRARY.

kartikdutt18: Hey @Yashwants19, I'll just push a commit. Thanks a lot.

tests/CMakeLists.txt
```diff
@@ -0,0 +1,58 @@
...
+               utils_tests.cpp
+)
+
+# Link dependencies of test executable.
+target_link_libraries(models_test
+  ${COMPILER_SUPPORT_LIBRARIES}
+  ${ARMADILLO_LIBRARIES}
+  ${BOOST_LIBRARIES}
+  ${Boost_FILESYSTEM_LIBRARY}
+  ${Boost_UNIT_TEST_FRAMEWORK_LIBRARY}
```
Yashwants19, line 20: Hi @kartikdutt18 do we have to include all boost libraries here again after including `BOOST_LIBRARIES`?

kartikdutt18: Yes, that was needed. It doesn't work without them.

Yashwants19: We can clear the list of boost libraries [here](https://github.com/mlpack/models/blob/de580de972993e54f4ff6f7a195ce384f33f5eb7/CMakeLists.txt#L183). It can be handled by visual studio itself.

.ci/windows-steps.yaml
kartikdutt18: Thanks a lot @zoq, @Yashwants19, The build works fine. To fix the architecture issue I passed Win64 flag and that fixed it. For the other error about mlpack was resolved by the above comments and the rest were related to installing some more boost libs.

Thanks a lot!
Regards.

kartikdutt18: Just cleaning commits now. Thanks a lot.

kartikdutt18:
Awesome, The build passed.
Hey @zoq, @saksham189, when you get a chance could you please take another look at this one.
Thanks a lot.

kartikdutt18:
Great, Thanks a lot for the review and the help.

kartikdutt18:
Awesome, Thanks a lot everyone!
For the reviews and amazing help. Thanks a lot.
I'll open a PR with documentation and LeNet (separate PRs) first thing tomorrow.
Regards.

PR #6, opened 2020-05-08
Adding BSD license in mlpack model repo

shrit:
Signed-off-by: Omar Shrit <shrit@lri.fr>

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -1,0 +1,29 @@
+BSD 3-Clause License
+
+Copyright (c) 2007-2020, mlpack
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are met:
+
+1. Redistributions of source code must retain the above copyright notice, this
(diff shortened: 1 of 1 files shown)


LICENSE.txt
```diff
@@ -0,0 +1,29 @@
+BSD 3-Clause License
+
+Copyright (c) 2007-2020, mlpack
```
zoq, line 3: Hm, not sure we can use the same year here, because the models repo is somewhat new; and this reads like we out a 'new' license to an older version, but I'm not an expert.

birm: That's a good question. I think it's reasonable to say that all of mlpack is split from one project, and thus can share the 2007 copyright date, but you're right that it probably makes more sense to use a newer year for this repo.

birm: Following the date on https://github.com/mlpack/models/commit/2f547f3345ca308f43b6d6a2d17fcdf94677f59f
```suggestion
Copyright (c) 2017-2020, mlpack
```

PR #16, opened 2020-06-04
transformer

mrityunjay-tripathi:
Hello everyone! I've implemented the transformer encoder and decoder. Though there are other dependencies for this PR, I made this PR to get some insights and opinions. Some things still remaining regarding this PR:
- [x] Get PR https://github.com/mlpack/mlpack/pull/2375 merged.
- [x] Implement positional encoding
- [ ] Adding wiki-dataset2
- [x] Adding tests for encoder and decoder
- [ ] Adding documentation for encoder and decoder.
- [ ] Do something useful using this code.

### DIFF
--- a/models/transformer/decoder.hpp
+++ b/models/transformer/decoder.hpp
@@ -60,7 +60,7 @@ class TransformerDecoder
    * @param srcSeqLen Source Sequence Length.
    * @param memoryModule The last Encoder module.
    * @param dModel The number of features in the input. Also, same as the
-   *        'embedDim' in 'MultiheadAttention' layer.
+   *        `embedDim` in `MultiheadAttention` layer.
    * @param numHeads The number of attention heads.
    * @param dimFFN The dimentionality of feedforward network.
    * @param dropout The dropout rate.
@@ -111,7 +111,7 @@ class TransformerDecoder
   /**
    * This method adds the attention block to the decoder.
    */
-  void AttentionBlock()
+  Sequential<>* AttentionBlock()
   {
     Sequential<>* decoderBlockBottom = new Sequential<>();
     decoderBlockBottom->Add<Subview<>>(1, 0, dModel * tgtSeqLen - 1, 0, -1);
@@ -173,14 +173,16 @@ class TransformerDecoder
     residualAdd2->Add(encoderDecoderAttention);
     residualAdd2->Add<IdentityLayer<>>();
 
-    decoder->Add(residualAdd2);
-    decoder->Add(new LayerNorm<>(dModel * tgtSeqLen));
+    Sequential<>* decoderBlock = new Sequential<>();
+    decoderBlock->Add(residualAdd2);
+    decoderBlock->Add<LayerNorm<>>(dModel * tgtSeqLen);
+    return decoderBlock;
   }
 
   /**
    * This method adds the position-wise feed forward network to the decoder.
    */
-  void PositionWiseFFNBlock()
+  Sequential<>* PositionWiseFFNBlock()
   {
     Sequential<>* positionWiseFFN = new Sequential<>();
     positionWiseFFN->Add<Linear3D<>>(dModel, dimFFN);
@@ -192,7 +194,10 @@ class TransformerDecoder
     AddMerge<>* residualAdd = new AddMerge<>();
     residualAdd->Add(positionWiseFFN);
     residualAdd->Add<IdentityLayer<>>();
-    decoder->Add(residualAdd);
+
+    Sequential<>* decoderBlock = new Sequential<>();
+    decoderBlock->Add(residualAdd);
+    return decoderBlock;
   }
 
   //! Locally-stored number of decoder layers.
--- a/models/transformer/decoder.hpp
+++ b/models/transformer/decoder.hpp
@@ -1,0 +1,236 @@
+/**
+ * @file models/transformer/decoder.hpp
+ * @author Mikhail Lozhnikov
+ * @author Mrityunjay Tripathi
+ *
+ * Definition of the Transformer Decoder layer.
+ *
+ * mlpack is free software; you may redistribute it and/or modify it under the
+ * terms of the 3-clause BSD license.  You should have received a copy of the
+ * 3-clause BSD license along with mlpack.  If not, see
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ */
+
+#ifndef MODELS_TRANSFORMER_DECODER_HPP
+#define MODELS_TRANSFORMER_DECODER_HPP
+
+#include <mlpack/prereqs.hpp>
+#include <mlpack/methods/ann/layer/layer_types.hpp>
+#include <mlpack/methods/ann/layer/base_layer.hpp>
+#include <mlpack/methods/ann/regularizer/no_regularizer.hpp>
+
+namespace mlpack {
+namespace ann /** Artificial Neural Network. */ {
+
+/**
+ * In addition to the two sub-layers in each encoder layer, the decoder inserts
+ * a third sub-layer, which performs multi-head attention over the output of the
+ * encoder stack. Similar to the encoder, we employ residual connections around
+ * each of the sub-layers, followed by layer normalization. We also modify the
+ * self-attention sub-layer in the decoder stack to prevent positions from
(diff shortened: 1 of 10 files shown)


kartikdutt18:
Hey @mrityunjay-tripathi, Thanks for opening this PR. If possible could you create a models folder and place the transformer inside. As many models will be added to the repo it would be great if we had a models folder rather than having a lot of models in main folder.

mrityunjay-tripathi:
Sure @kartikdutt18! Actually I was also thinking about why it was not that way. But no worries. I will make it that way now :+1:

kartikdutt18:
> Sure @kartikdutt18! Actually I was also thinking about why it was not that way. But no worries. I will make it that way now 👍

Awesome, The reason for that it is part of Restructuring - 3. All existing models will be replaced with something similar to what you have implemented i.e. in a class so that user could use pre trained models.

models/transformer/decoder.hpp
```diff
@@ -0,0 +1,150 @@
...
+  /**
+   * Create the TransformerDecoder object using the specified parameters.
+   *
+   * @param dModel The number of features in the input. Also, same as the
+   *        `embedDim` in `MultiheadAttention` layer.
+   * @param numHeads The number of attention heads.
+   * @param dimFFN The dimentionality of feedforward network.
+   * @param dropout The dropout rate.
+   */
+  TransformerDecoder(const size_t dModel = 512,
```
zoq, line 65: Does d in Model stand for dimension? Maybe we can find a more descriptive name, e.g. `modelDimension`  or `numFeatures`?

models/transformer/decoder_impl.hpp
```diff
@@ -0,0 +1,115 @@
...
+  ffnBlock.Add<Linear<InputDataType, OutputDataType, RegularizerType> >(dModel, dimFFN);
+  ffnBlock.Add<ActivationFunction>();
+  ffnBlock.Add<Linear<InputDataType, OutputDataType, RegularizerType> >(dimFFN, dModel);
+  ffnBlock.Add<Dropout<InputDataType, OutputDataType> >(dropout);
+}
+
+template <typename InputDataType, typename OutputDataType,
+    class ActivationFunction, typename RegularizerType>
+void TransformerDecoder<InputDataType, OutputDataType, ActivationFunction,
+RegularizerType>::Assemble()
```
zoq, line 62: Most probably I missed something, so I guess there is a good reason to not put this into the constructor?

mrityunjay-tripathi: Ahh, yes. Because the `Concatenate` layer requires the key matrix (line 64) to be concatenated (line 67 & 68) and how can one assign `key` if object is not yet constructed? (`key` is the output of the Encoder). Though it was not required for `Encoder`, yet I made them separate to maintain the same API. The name of the method might not be a good choice 😁, so I am open to any other name for this method. When I named it, in my mind - "Avengers! Assemble" :smiley:

models/transformer/decoder.hpp
mrityunjay-tripathi: Yup! Sure. Actually the [paper](https://arxiv.org/abs/1706.03762) used "d<sub>model</sub>" everywhere. As mentioned in the argument description, `dModel` is same as `embedDim` (size of each embedding vector). Should I make it `embedDim` (or `embeddingDim` or `embeddingDimension`) as it says more about what actually it is.

models/transformer/decoder_impl.hpp
zoq: Ahh, thanks for the clarification, I like the name.

models/transformer/decoder_impl.hpp
```diff
@@ -0,0 +1,129 @@
...
+
+  network.Add(attnBlock1);
+  network.Add<LayerNorm<InputDataType, OutputDataType> >(dModel * tgtSeqLen);
+
+  // if (key.is_empty())
+  //   Log::Fatal << "The 'key' matrix is required but empty!" << std::endl;
+
+  // MultiheadAttention<InputDataType, OutputDataType, RegularizerType>
+  //     encoderDecoderAttn(dModel, numHeads);
+  // encoderDecoderAttn.Key() = key;
```
lozhnikov, line 72: Regarding your question. If I'm not mistaken, you need to concatenate the output of the Encoder with the previous layer and pass it to `encoderDecoderAttn`. But don't forget the Encoder and the Decoder have different inputs. You need to split a dataset record into two parts: `(input, target)`. You can use the `Subview` layer for that. I tried to implement a rough model [here](https://github.com/mlpack/mlpack/pull/2375#issuecomment-672910701). I think it should provide a clue on your question.

Tell me if you have any questions. It's not that easy.

mrityunjay-tripathi: The problem is neither we can pass more than one inputs to `FFN` nor to the `Concat` layer. I can understand the use of Concat for repeating query into `[query query query]` but how will we use it to get `[query key key]` or 
`[query key value]`? I think defining the `Backward` and `Gradient` function manually is not a bad option (It will use less number of layers also). Like if we define single encoder and single decoder block as a layer. Then the complete encoder could be `N` deep copies of the encoder block and similarly for the complete decoder. 
I think I'm not understanding what is going on in your mind :( Could you please clarify how two inputs are flowing in the `encoderDecoderAttn` part? And how their backpropagated errors will reach to appropriate place?

lozhnikov: > The problem is neither we can pass more than one inputs to FFN nor to the Concat layer.

Of course we can. Technically we pass jest one input to each layer. But we can use the `Subview` layer in order to split the input in two parts. Then we'll process each input separately. Look at the example https://github.com/mlpack/mlpack/pull/2375#issuecomment-672910701.

> Could you please clarify how two inputs are flowing in the encoderDecoderAttn part?

I think the example https://github.com/mlpack/mlpack/pull/2375#issuecomment-672910701 should clarify that.

> I think defining the Backward and Gradient function manually is not a bad option (It will use less number of layers also)

Yes, the backward propagation will require less number of layers. But the implementation itself would be more complex.

mrityunjay-tripathi:
I was trying to this test locally and I got following error--
```sh
error: matrix multiplication: incompatible matrix dimensions: 0x0 and 16x10
unknown location(0): fatal error: in "FFNModelsTests/TransformerEncoderTest": std::logic_error: matrix multiplication: incompatible matrix dimensions: 0x0 and 16x10
```
I feel there is some problem with Reset. The weights and biases are not allocated memory. But I can't find why.

lozhnikov:
> I feel there is some problem with Reset. The weights and biases are not allocated memory. But I can't find why.

I'll look into it in the morning.

lozhnikov:
@mrityunjay-tripathi I think I get it. Looks like my comment https://github.com/mlpack/models/pull/16#discussion_r475212736 was wrong. We need to pass `model = true` to the constructors (I mean `Sequential`, `Concat`, `AddMerge`). In this case all the visitors will go through their sublayers. And there will be no memory issues since the Delete visitor do the same.

mrityunjay-tripathi:
>We need to pass model = true to the constructors (I mean Sequential, Concat, AddMerge).

Yeah. Got it. Thanks :)

lozhnikov:
@mrityunjay-tripathi Is this PR ready? Can I review this?

mrityunjay-tripathi:
Yes. This is ready for review.

models/transformer/decoder.hpp
```diff
@@ -0,0 +1,263 @@
...
+                                                          srcSeqLen,
+                                                          dModel,
+                                                          numHeads);
+    mha2->KeyPaddingMask() = keyPaddingMask;
+    encoderDecoderAttention->Add(mha2);
+
+    // Residual connection.
+    AddMerge<>* residualAdd2 = new AddMerge<>(true);
+    residualAdd2->Add(encoderDecoderAttention);
+    residualAdd2->Add(decoderBlockBottom);
```
lozhnikov, line 194: You can't pass the same block twice (see the comment to `encDecAttnInput`). Looks like we need to change the model a bit. I have to go now. I'll come up with the idea in the evening.

lozhnikov (commented):
Sorry for the slow response.  The beginning of the term was hard. Now things settled a bit. I found a tiny flaw in the decoder implementation. I'll suggest the fix in the evening.

PR #25, opened 2020-07-11
Add YOLO PreProcessor

kartikdutt18:
Hey everyone,
This PR aims to add preprocessor function for YOLO class. This basically converts annotations into feature maps that can be used for training YOLO model i.e. have the same format as output of YOLO model.
To Do :
- [x] Add tests to verify output.

CC : @KimSangYeon-DGU, @saksham189, @zoq

### DIFF
--- a/dataloader/preprocessor.hpp
+++ b/dataloader/preprocessor.hpp
@@ -194,15 +194,19 @@ class PreProcessor
 
         if (normalize)
         {
-          gridCoordinates(0) = std::ceil(gridCoordinates(0) / cellSizeWidth) - 1;
-          gridCoordinates(1) = std::ceil(gridCoordinates(1) / cellSizeHeight) - 1;
+          gridCoordinates(0) = std::ceil(gridCoordinates(0) /
+              cellSizeWidth) - 1;
+          gridCoordinates(1) = std::ceil(gridCoordinates(1) /
+              cellSizeHeight) - 1;
         }
         else
         {
-          gridCoordinates(0) = std::ceil((gridCoordinates(0) / imageWidth) / cellSizeWidth) - 1;
-          gridCoordinates(1) = std::ceil((gridCoordinates(1) / imageHeight) / cellSizeHeight) - 1;
+          gridCoordinates(0) = std::ceil((gridCoordinates(0) /
+              imageWidth) / cellSizeWidth) - 1;
+          gridCoordinates(1) = std::ceil((gridCoordinates(1) /
+              imageHeight) / cellSizeHeight) - 1;
         }
-        
+
 
         size_t gridX = gridCoordinates(0);
         size_t gridY = gridCoordinates(1);
--- a/dataloader/preprocessor.hpp
+++ b/dataloader/preprocessor.hpp
@@ -162,9 +162,9 @@ class PreProcessor
             (i + 1) * 5 - 1));
       }
 
-      // For YOLOv2 or higher, each bounding box can represent a class so we don't
-      // Repeat labels as done for YOLOv1. We will use map to store last inserted
-      // bounding box.
+      // For YOLOv2 or higher, each bounding box can represent a class
+      //  so we don't repeat labels as done for YOLOv1. We will use map
+      //  to store last inserted bounding box.
       std::map<std::pair<size_t, size_t>, size_t> boundingBoxOffset;
 
       // Normalize the coordinates.
@@ -232,7 +232,8 @@ class PreProcessor
           outputTemp(arma::span(gridX), arma::span(gridY),
               arma::span(bBoxOffset, bBoxOffset + 1)) = gridCoordinates;
           outputTemp(arma::span(gridX), arma::span(gridY),
-              arma::span(bBoxOffset + 2, bBoxOffset + 3)) = widthAndHeight.col(i);
+              arma::span(bBoxOffset + 2,
+                  bBoxOffset + 3)) = widthAndHeight.col(i);
           outputTemp(gridX, gridY, bBoxOffset + 4) = 1.0;
           outputTemp(gridX, gridY, bBoxOffset + 5 + labels.col(i)(0)) = 1;
         }
--- a/dataloader/preprocessor.hpp
+++ b/dataloader/preprocessor.hpp
@@ -100,7 +100,7 @@ class PreProcessor
    * the size : gridWidth * gridHeight * (5 * numBoxes + classes).
    *
    * @param annotations Field object created using model's dataloader containing
-   *                    annotation for images.
+   *     annotation for images.
    * @param output Output matrix where output will be stored.
    * @param imageWidth Width of image used for training YOLO model.
    * @param imageHeight Height of image used for training YOLO model.
@@ -152,7 +152,7 @@ class PreProcessor
     for (size_t boxIdx = 0; boxIdx < batchSize; boxIdx++)
     {
       arma::cube outputTemp(const_cast<arma::Mat<eT> &>(output).memptr() +
-        offset, gridHeight, gridWidth, numPredictions, false, false);
+          offset, gridHeight, gridWidth, numPredictions, false, false);
       offset += gridWidth * gridHeight * numPredictions;
 
       // Get the bounding box and labels corresponding to current image.
@@ -166,8 +166,8 @@ class PreProcessor
       }
 
       // For YOLOv2 or higher, each bounding box can represent a class
-      //  so we don't repeat labels as done for YOLOv1. We will use map
-      //  to store last inserted bounding box.
+      // so we don't repeat labels as done for YOLOv1. We will use map
+      // to store last inserted bounding box.
       std::map<std::pair<size_t, size_t>, size_t> boundingBoxOffset;
 
       // Normalize the coordinates.
@@ -206,7 +206,6 @@ class PreProcessor
           gridCoordinates(1) = std::ceil((gridCoordinates(1) /
               imageHeight) / cellSizeHeight) - 1;
         }
-
 
         size_t gridX = gridCoordinates(0);
         size_t gridY = gridCoordinates(1);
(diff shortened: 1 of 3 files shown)


dataloader/preprocessor.hpp
```diff
@@ -48,12 +48,112 @@ class PreProcessor
...
+          annotations(0, boxIdx).n_rows - 1), arma::span());
+      arma::vec labels = annotations(0, boxIdx).row(0);
+
+      // Normalize the coordinates.
+      boundingBoxes.row(0) /= imageWidth;
+      boundingBoxes.row(2) /= imageWidth;
+      boundingBoxes.row(1) /= imageHeight;
+      boundingBoxes.row(3) /= imageHeight;
+
+      // Get width and height as well as centres for the bounding box.
```
zoq, line 118: Should this be "[...] as well as the center for the bounding box."?

kartikdutt18: Right, I will fix it in the next commit.

dataloader/preprocessor.hpp
```diff
@@ -48,12 +48,111 @@ class PreProcessor
...
+      // Get the bounding box and labels corresponding to current image.
+      arma::mat boundingBoxes = annotations(0, boxIdx)(arma::span(1,
+          annotations(0, boxIdx).n_rows - 1), arma::span());
+      arma::vec labels = annotations(0, boxIdx).row(0);
+
+      // Normalize the coordinates.
+      boundingBoxes.row(0) /= imageWidth;
+      boundingBoxes.row(2) /= imageWidth;
+      boundingBoxes.row(1) /= imageHeight;
+      boundingBoxes.row(3) /= imageHeight;
```
KimSangYeon-DGU, line 115: Hey @kartikdutt18 
To normalize the width and height of image into -1 and 1, can you refer to [this](https://stats.stackexchange.com/questions/178626/how-to-normalize-data-between-1-and-1)?

KimSangYeon-DGU: Ahh, I misunderstood the question

saksham189:
Hey @kartikdutt18 , I just wanted to know whether you were stuck anywhere with this because we were planning to finish this first.

kartikdutt18:
Hey @saksham189, I've probably written some expression wrong somewhere and I wasn't able to track it down before Saturday so as I implemented the weight translator [here](https://github.com/kartikdutt18/mlpack-PyTorch-Weight-Translator) so we can now transfer weights of Darknet model or any other model from PyTorch to mlpack. However other than weights we also need to transfer running mean and variance for BatchNorm Layer and with that I'm stuck with the error [here](https://github.com/mlpack/mlpack/pull/2519#issuecomment-661120039). I'll take a look at the preprocessor again today. In the meanwhile, I tried training Darknet 19 model on Cifar10, It didn't converge, so I also added Darknet 1 (same as YOLO model) on Cifar10 to see if we can train that and it takes 3 hours for a single epoch so we should also be able to train YOLO model (in less time than Darknet), if needed. Here is the log for the first epoch (20% accuracy).
21_log_epoch

saksham189:
Looks good and the training time is also considerably lower. So, maybe we should just train that first and look at the accuracy.

kartikdutt18:
> Looks good and the training time is also considerably lower. So, maybe we should just train that first and look at the accuracy.

Great, I'm letting the model train. In the meanwhile, I will try to get running mean and variance also loaded into mlpack. Also, with transferred weights we can also train the model on imagenette (that way running mean and variance will be set in training).

kartikdutt18:
Hey @saksham189, I fixed the errors in the implementation as well added tests fro single image, multiple images and image with multiple bounding box.

tests/preprocessor_tests.cpp
```diff
@@ -0,0 +1,61 @@
...
+  bBox << 7 << 157 << 90 << 486 << 372 << arma::endr;
+  input(0, 2) = bBox;
+
+  PreProcessor<arma::mat, arma::field<arma::vec>>::YOLOPreProcessor(
+      input, output, 500, 387);
+
+  arma::vec desiredSum(3);
+  desiredSum << 8.3342 << 18.4093 << 7.13195 << arma::endr;
+  for (size_t i = 0; i < output.n_cols; i++)
+    BOOST_REQUIRE_CLOSE(arma::accu(output.col(i)), desiredSum(i), 1e-3);
```
zoq, line 58: Maybe we should also compare the normalized boxes with pre calculated ones, what do you think?

dataloader/preprocessor.hpp
```diff
@@ -48,12 +48,117 @@ class PreProcessor
...
+      // Get the bounding box and labels corresponding to current image.
+      arma::mat labels(1, annotations(0, boxIdx).n_elem / 5);
+      arma::mat boundingBoxes(4, annotations(0, boxIdx).n_elem / 5);
+      for (size_t i = 0; i < boundingBoxes.n_cols; i++)
+      {
+        labels.col(i)(0) = annotations(0, boxIdx)(i * 5);
+        boundingBoxes.col(i) = annotations(0, boxIdx)(arma::span(i * 5 + 1,
+            (i + 1) * 5 - 1));
+      }
+
```
zoq, line 115:
```suggestion
```

Looks like we can remove the extra line here.

dataloader/preprocessor.hpp
```diff
@@ -48,12 +48,117 @@ class PreProcessor
...
+
+      // Normalize the coordinates.
+      boundingBoxes.row(0) /= imageWidth;
+      boundingBoxes.row(2) /= imageWidth;
+      boundingBoxes.row(1) /= imageHeight;
+      boundingBoxes.row(3) /= imageHeight;
+
+      // Get width and height as well as centres for the bounding box.
+      arma::mat widthAndHeight(2, boundingBoxes.n_cols);
```
zoq, line 124: I have seen at least one dataset that alreay used the centered bbox format but wasn't normalized, do you think we should introduce a parameter to skip the step?

tests/preprocessor_tests.cpp
kartikdutt18: Sure, I can add that as well.

dataloader/preprocessor.hpp
kartikdutt18: Makes sense. Thanks for the suggestion.

dataloader/preprocessor.hpp
kartikdutt18: Ahh makes sense. I'll add that as well.

kartikdutt18:
>  So, maybe we should just train that first and look at the accuracy.

Hey @saksham189, @KimSangYeon-DGU, I continued the training. Initially it was for 5 epochs (the learning rate was 0.01) and after 5 epochs it remained constant and the below graph was obtained. Then,  I set learning rate to 0.001 and it's still training and has completed nearly 25 epochs out of 40. The training and validation accuracy are continuously increasing from 20% and 7% respectively to 31% and roughly 20%. The graph for that is also attached below. Also, about using larger image sizes such as imagenette, The training time doesn't change significantly with larger images however time for loading the dataset is very large. Imagenette with 320 x 320 x 3 image size and 5k images nearly took an hour.
**Current training log (lr = 1e-3).**
darknet_v1_fast_mean_log_epoch
darknet_v1_fast
 
**First attempt to train (5 epochs / lr = 1e-2) (training accuracy = 26%, Graphs are shown below)**
darknet_v1_slow_mean_log_epoch
darknet_v1_slow
I'll add the changes for YOLOv3 in a couple of minutes. Then I would also have to add tests for the same.

dataloader/preprocessor.hpp
```diff
@@ -93,6 +93,174 @@ class PreProcessor
...
+                               const size_t imageHeight = 224,
+                               const size_t gridWidth = 7,
+                               const size_t gridHeight = 7,
+                               const size_t numBoxes = 2,
+                               const size_t numClasses = 20,
+                               const bool normalize = true)
+  {
+    // See if we can change this to v4 / v5.
+    mlpack::Log::Assert(version >= 1 && version <= 3, "Supported YOLO versions \
+        are version 1 to version 3.");
```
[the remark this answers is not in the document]
kartikdutt18: Yes it does. It might also work for YOLOv4 unless they changed feature map but it works for all versions between 1 and 3 (inclusive).

kartikdutt18:
I think this is ready to go as well.

dataloader/preprocessor.hpp
```diff
@@ -93,6 +93,173 @@ class PreProcessor
...
   }
+
+  /**
+   * PreProcessor for YOLO model. Converts arma::field type annotations to
+   * arma::mat type for training YOLO model. Each column in target matrix has
+   * the size : gridWidth * gridHeight * (5 * numBoxes + classes).
+   *
+   * @param annotations Field object created using model's dataloader containing
+   *     annotation for images.
+   * @param output Output matrix where output will be stored.
```
KimSangYeon-DGU, line 104: Hey @kartikdutt18, can you explain the input, output, and terms of `cell` and `grid` which are used in this implementation?

KimSangYeon-DGU (commented):
Sorry for the late response. I left a question for me to get understood this feature.

dataloader/preprocessor.hpp
kartikdutt18: Hey @KimSangYeon-DGU, no worries, The input here would be a field type containing bounding boxes, the output would be an arma::mat which be would be of size batchSize * (featureMapSize) where featureSize would correspond to  S * S * (numBoundingBoxes * (numClasses + 5)) for YOLOv3 and S * S * (numBoundingBoxes * 5 + numClasses). Here cell is a single cell / data point in Feature Size. And the grid is featureSize reshaped as a cube (S, S, (numBoundingBoxes * 5 + numClasses)).

KimSangYeon-DGU (approved):
Looks good to me.

saksham189 (approved):

saksham189:
Thanks for the great contribution @kartikdutt18 !

kartikdutt18:
Thanks  lot @KimSangYeon-DGU, @saksham189, @zoq for the reviews and all the help. Also, thanks a lot for mentoring me during GSoC.

PR #32, opened 2020-08-20
[WIP] YOLOv3 Layer

kartikdutt18:
Hey everyone,
This PR aims to add YOLOv3 model. Currently I have added a generalized of forward propogation of YOLOv3 layer which should work for both YOLOv3 and tiny YOLOv3. We would only have to pass the DarkNet 53 blocks from DarkNet PR into this and it will form the necessary connections.
Regards,
Kartik.


kartikdutt18:
Keep Open.

PR #34, opened 2020-08-31
Fix azure build.

kartikdutt18:
Hey everyone,
This PR aims to fix the azure build for windows. The required libraries for R aren't installed so it causes an build error.
Regards,
Kartik.


rcurtin:
Thanks @kartikdutt18!

kartikdutt18:
Great, Thanks a lot @rcurtin for the review.

PR #47, opened 2021-02-20
Fix build.

### LINKED ISSUE #46: Fix build.
With recent changes in mlpack such as switches in serialisation etc. The azure build is failing hence the build needs to be fixed.

kartikdutt18:
closes #46.


.ci/linux-steps.yaml
```diff
@@ -7,6 +7,11 @@ steps:
 # Install Build Dependencies.
 # Install mlpack from master to get new functionalities instantly.
 - script: |
+    # Install cereal.
+    wget https://github.com/USCiLab/cereal/archive/v1.3.0.tar.gz
+    tar -xvzpf v1.3.0.tar.gz # Unpack into cereal-1.3.0/.
+    cd cereal-1.3.0/
```
rcurtin, line 13: Maybe this `cd` isn't necessary?

zoq: Nice catch, this should hopefully also fix the linux build.

zoq (commented):
Not sure yet, why it finds armadillo but fails afterwards: 

```
-- Found Armadillo: /usr/local/lib/libarmadillo.dylib (found suitable version "10.2.1", minimum required is "8.400.0") 
CMake Error at CMakeLists.txt:114 (find_package):
  find_package called with invalid argument ""
```

rcurtin:
Looks like everything is building fine, but maybe the URLs are wrong in `utils_test.cpp`?  Let me know if you need me to add any files to mlpack.org. :+1:

kartikdutt18:
Hey @rcurtin, It would be great if you could upload the datasets in this [link](https://drive.google.com/drive/folders/1gi9Tj_8lJCYpFba7UX_-YAJDSTqQwBIE?usp=sharing) to mlpack.org/datasets. This should fix the build and not break the datasets link in the repo.

rcurtin:
I think those are already there:

http://mlpack.org/datasets/iris.csv
http://mlpack.org/datasets/mnist.tar.gz
http://mlpack.org/datasets/PASCAL-VOC-Test.tar.gz
http://mlpack.org/datasets/USCensus1990.tar.gz

There was a bug where the correct version of the datasets was not being downloaded, so if those builds are more than ~a week old, that may be the issue here and maybe we just need to re-run.

kartikdutt18:
Awesome, the build is passing now!

CMake/Findcereal.cmake
```diff
@@ -0,0 +1,56 @@
+#Findcereal.cmake
```
birm, line 1: Thought of sadding this to https://gitlab.kitware.com/cmake/cmake/-/tree/master/Modules ? :smiley:

rcurtin (approved):
Awesome, great to see the build passes now! :+1:

zoq (approved):
Nice :+1:

kartikdutt18:
Great, Thanks everyone for the reviews and suggestions.

CMake/Findcereal.cmake
kartikdutt18: I will take a look at this as well. Thanks!

PR #55, opened 2021-04-23
Namespace and include guard 

### LINKED ISSUE #54: Namespace and Include guards
Points to Discuss:

1. For now the models repo uses mlpack namespace in some files and just plain nothing in some so should this repository have its namespace in all the files or just carry the mlpack namespace.
2. the include guard in all the files follows a format as MODELS_FILENAME.HPP, I propose that it follows a similar structure such as mlpack which will be MODELS_FOLDER_NAME_FILENAME.HPP.

also please keep in mind that for now this repository doesn't have any installation candidate or place that it is installed in but that will be included in it very soon, I am just figuring out how to configure install with cmake and i will do that then.

Aakash-kaushik:
Implements issues discussed in #54 
Please merge it only after #52 
closes #54


Aakash-kaushik:
I have one more proposal, I think it is better to keep the model hpp files directly inside the models folder in the repo rather than having each individual folder. this will allow the user to do something like this:
```c++
#include <models/models/darknet.hpp>
``` 
rather than
```c++
#include <models/models/darknet/darknet.hpp>
```

.ci/windows-steps.yaml
```diff
@@ -10,7 +10,6 @@ steps:
 - powershell: |
     nuget install OpenBLAS -o $(Agent.ToolsDirectory)
     nuget install boost -o $(Agent.ToolsDirectory) -Version 1.60.0
-    nuget install boost_unit_test_framework-vc140 -o $(Agent.ToolsDirectory) -Version 1.60.0
     nuget install boost_program_options-vc140 -o $(Agent.ToolsDirectory) -Version 1.60.0
```
rcurtin, line 13: Nice to remove that!  I think we could also remove boost::program_options and boost::serialization too, unless the models repository is using them.  At the very least, mlpack itself is only using cereal and CLI11 (packaged internally).

CMakeLists.txt
```diff
@@ -170,7 +170,6 @@ find_package(Boost 1.49
       regex
       program_options
       serialization
```
rcurtin, line 172:
```suggestion
```
I think you should be able to remove these too. :+1:

models/darknet/darknet.hpp
```diff
@@ -47,7 +47,8 @@
 #include <mlpack/methods/ann/init_rules/glorot_init.hpp>
 
 namespace mlpack {
-namespace ann /** Artificial Neural Network. */{
+namespace ann /* Artificial neural networks */{
+namespace models {
```
rcurtin, line 51: Personally I think it might be cleaner to put the various networks in `mlpack::models`, instead of `mlpack::ann::models`.  Let me know what you think. :+1:

tests/CMakeLists.txt
```diff
@@ -19,7 +20,6 @@ target_link_libraries(models_test
   ${COMPILER_SUPPORT_LIBRARIES}
   ${ARMADILLO_LIBRARIES}
   ${Boost_FILESYSTEM_LIBRARY}
-  ${Boost_UNIT_TEST_FRAMEWORK_LIBRARY}
   ${Boost_SYSTEM_LIBRARY}
   ${Boost_SERIALIZATION_LIBRARY}
```
rcurtin, line 24:
```suggestion
```
Here's another thing we could remove.  I think in these suggestions I haven't caught everything, so maybe there is more to do too.

tests/CMakeLists.txt
```diff
@@ -30,32 +30,7 @@ target_link_libraries(models_test
 # Get the list of sources from the test target.
 get_target_property(test_sources models_test SOURCES)
 
-# Go through the list of test sources and parse the test suite name.
```
rcurtin, line 33: Should we really remove this?  If we transitioned to using catch, then I think we can regex on `TEST_CASE(..., <test suite name>)` instead of `BOOST_AUTO_TEST_SUITE` and it should still work fine.  It might take a little bit of work to adapt it.

tests/dataloader_tests.cpp
```diff
@@ -9,74 +9,77 @@
...
-#define BOOST_TEST_MAIN // Do not define this anywhere else.
-#include <dataloader/dataloader.hpp>
-#include <boost/test/unit_test.hpp>
 
-using namespace boost::unit_test;
+/**
+ * CATCH_CONFIG_MAIN tells Catch to provide a main() don't include
+ * this in other test files.
+ */
+#define CATCH_CONFIG_MAIN
```
rcurtin, line 17: I'd suggest defining `main()` ourselves so that we can, e.g., set the random seed.  Take a look at how it's done in ensmallen: https://github.com/mlpack/ensmallen/blob/master/tests/main.cpp

rcurtin (commented):
Very nice work @Aakash-kaushik!  I have a few minor comments. :+1:

models/darknet/darknet.hpp
Aakash-kaushik: So just having mlpack::models was creating a issue that i cannot just use the ann namespace inside the namespace by doing the usual `using namespace mlpack::ann` and if i can't do that i have to go everywhere and use the the whole naming convention like `mlpack::ann::component_i_want_to_use` so to avoid huge code changes I used the ann namespace too. 
Is there a workaround?

tests/CMakeLists.txt
Aakash-kaushik: I mean yes we can adapt it to catch but i don't know what purpose it serves ?

tests/dataloader_tests.cpp
Aakash-kaushik: I have applied this suggestion, have a look.

zoq:
Let's see if I fixed the issue.

zoq:
@mlpack-jenkins test this please

zoq (commented):
Looks like this include some changes from the merged catch transition PR. Can you merge this PR against the latest master branch.

Aakash-kaushik:
> Let's see if I fixed the issue.

Man you are magic, what did you do ?
I don't know how jenkins and style checks works but would you mind explaining what the problem was, I am curious.

Aakash-kaushik:
> Looks like this include some changes from the merged catch transition PR. Can you merge this PR against the latest master branch.

Yes it does but i thought as soon as the changes from that catch PR are merged the files in this PR would dynamically update, but that doesn't seems to be the case.

zoq:
> > Let's see if I fixed the issue.
> 
> Man you are magic, what did you do ?
> I don't know how jenkins and style checks works but would you mind explaining what the problem was, I am curious.

The Jenkins Style Check job used an outdated plugin (GitHub updated the API) to push the job status to GitHub, we already used another plugin that handles the communication with the GitHub API, so that outdated plugin isn't needed anymore.

models/darknet/darknet.hpp
rcurtin: I think that just `ann::component_i_want_to_use` should work too, since both are under the `mlpack` namespace.  But still, I see your point.  The main reason I suggested `mlpack::models` instead of `mlpack::ann::models` is that someday we might add models to this repository that are not ANN models, and so the `mlpack::ann::models` namespace might not make sense.  (Of course, in that case, we could just use `mlpack::models` there, perhaps?)

I can see advantages and disadvantages to both approaches, but I'm not sure which is better.  From a user perspective, `mlpack::models` is probably easiest and clearest, but from a developer perspective, the disadvantage is that we may need `ann::` in the code.

What do you think?  Also, @kartikdutt18, @zoq, and others, any opinions or thoughts?  I'm pretty on the fence on this one; no strong opinion.

tests/CMakeLists.txt
rcurtin: Essentially this part of the code allows CTest integration for each model type.  So, with this code, we have the ability to have one CTest test suite for each model type, which is nice.  Otherwise, if you used CTest, it groups everything as one test suite, but especially for these models, tests for a model type may take a long time and so it might be preferable to have the ability to run only one test at a time.

rcurtin:
From my perspective everything looks great to merge once we handle the namespace and CTest issues. :+1:  Thanks for working on this @Aakash-kaushik!

models/darknet/darknet.hpp
Aakash-kaushik: I am on the side of `mlpack::model` because of these points 
1. There could be more models that are not ANN
2. Easy for user  

Let me know what others think and i can make the changes.

tests/CMakeLists.txt
Aakash-kaushik: Makes absolute sense, I will implement it.

Aakash-kaushik:
> From my perspective everything looks great to merge once we handle the namespace and CTest issues. 👍 Thanks for working on this @Aakash-kaushik!

Happy to learn and work on mlpack any day.

models/darknet/darknet.hpp
zoq: Agreed, the plan was always to use this for everything mlpack implements, the ann part is only a small part, so I think it makes sense if the namespace reflects that.

zoq:
> > Looks like this include some changes from the merged catch transition PR. Can you merge this PR against the latest master branch.
> 
> Hey @zoq i did merge this agains models/master but it still shows all the same changes.
> I can open a new PR but it would be hard to keep track of the comments and things that ryan left.

Opening a PR might be the easiest solution, since the PR is just going to be closed afterwards the comments are not lost and the PR is pretty close to get merged anyway.

models/darknet/darknet.hpp
Aakash-kaushik: i'll start making the changes.

Aakash-kaushik:
> > > Looks like this include some changes from the merged catch transition PR. Can you merge this PR against the latest master branch.
> > 
> > 
> > Hey @zoq i did merge this agains models/master but it still shows all the same changes.
> > I can open a new PR but it would be hard to keep track of the comments and things that ryan left.
> 
> Opening a PR might be the easiest solution, since the PR is just going to be closed afterwards the comments are not lost and the PR is pretty close to get merged anyway.

So i will close this PR when i open the new one with changes and will tag this PR in that too.

Aakash-kaushik:
Closing in favour of a new PR because it carried some commits from another PR and was comparing changes the wrong way.

PR #61, opened 2021-05-29
[WIP] Resnet Module

Aakash-kaushik:
This PR aims to implement resnet module which would be able to create all the resnet variants from the paper and this aims to follow the same architecture as PyTorch for some reasons.
1. We can't train so many models on imagenet right now.
2. We don't know if they will converge.
3. keeping the same architecture that allows us to get weights from PyTorch.

Things i have some doubts about: 
1. How would the residual block be implemented from the sequential layer?
2. Can I get the output of a layer at a random stage and add it to another layer? (skip connections)

Resources:
1. Pytorch's [resnet implementation](https://github.com/pytorch/vision/blob/master/torchvision/models/resnet.py).
2. [Resnet paper](https://arxiv.org/abs/1512.03385).


zoq:
The idea of the `sequential` layer is that it wraps arbitrary layers and exposes them as if it would be a single layer. The `sequential` layer has a template parameter (https://github.com/mlpack/mlpack/blob/83e70110595eaf3cf3758f270433801e673615b2/src/mlpack/methods/ann/layer/sequential.hpp#L70), which tells the layer to add the input to the output of the last layer. There is also a convenient typedef (https://github.com/mlpack/mlpack/blob/83e70110595eaf3cf3758f270433801e673615b2/src/mlpack/methods/ann/layer/sequential.hpp#L260-L261) that already sets the template parameter for you. Below is an elementary example:

```c++
 Residual<>* residual = new Residual<>(true);

Linear<>* linearA = new Linear<>(10, 10);
Linear<>* linearB = new Linear<>(10, 10);
	 
residual->Add(linearA);
residual->Add(linearB);
```

in this case `linearA` and `linearB` are run and the input is also added to the output of the last layer, which in this case is `linearB`.

There is also a test case - https://github.com/mlpack/mlpack/blob/83e70110595eaf3cf3758f270433801e673615b2/src/mlpack/tests/ann_layer_test.cpp#L3325

Aakash-kaushik:
> The idea of the `sequential` layer is that it wraps arbitrary layers and exposes them as if it would be a single layer. The `sequential` layer has a template parameter (https://github.com/mlpack/mlpack/blob/83e70110595eaf3cf3758f270433801e673615b2/src/mlpack/methods/ann/layer/sequential.hpp#L70), which tells the layer to add the input to the output of the last layer. There is also a convenient typedef (https://github.com/mlpack/mlpack/blob/83e70110595eaf3cf3758f270433801e673615b2/src/mlpack/methods/ann/layer/sequential.hpp#L260-L261) that already sets the template parameter for you. Below is an elementary example:
> 
> ```c++
>  Residual<>* residual = new Residual<>(true);
> 
> Linear<>* linearA = new Linear<>(10, 10);
> Linear<>* linearB = new Linear<>(10, 10);
> 	 
> residual->Add(linearA);
> residual->Add(linearB);
> ```
> 
> in this case `linearA` and `linearB` are run and the input is also added to the output of the last layer, which in this case is `linearB`.
> 
> There is also a test case - https://github.com/mlpack/mlpack/blob/83e70110595eaf3cf3758f270433801e673615b2/src/mlpack/tests/ann_layer_test.cpp#L3325

Hi thanks @zoq for this but the part that confused me was that it checks in the code that if the dimensions of the first layer are equal to the last layer or not and for ResNet there would be a case where the input dim of the first layer will be diff from the last one and so I need a conv 1*1 block just for the first layer input that is not run like all layers but separately before adding it to the last layer. 

how do you suggest i accomplish that ?

zoq:
In this case you can use a combination of `AddMerge` and `Sequential`. The `AddMerge`  layer just takes arbitrary runs each layer and at the end adds them together.

```c++
AddMerge<> resblock(false, false);

Sequential<>* sequential = new Sequential<>(true);

Linear<>* linearA = new Linear<>(10, 10);
Linear<>* linearB = new Linear<>(10, 10);

 sequential->Add(linearA);
 sequential->Add(linearB);

Convolution<> conv = new Convolution<>(...);

resblock.Add(sequential);
resblock.Add(conv);
```

Let me know if this is what you looked for. Maybe it makes sense to implement that structure as an independent layer in mlpack.

Aakash-kaushik:
> In this case you can use a combination of `AddMerge` and `Sequential`. The `AddMerge` layer just takes arbitrary runs each layer and at the end adds them together.
> 
> ```c++
> AddMerge<> resblock(false, false);
> 
> Sequential<>* sequential = new Sequential<>(true);
> 
> Linear<>* linearA = new Linear<>(10, 10);
> Linear<>* linearB = new Linear<>(10, 10);
> 
>  sequential->Add(linearA);
>  sequential->Add(linearB);
> 
> Convolution<> conv = new Convolution<>(...);
> 
> resblock.Add(sequential);
> resblock.Add(conv);
> ```
> 
> Let me know if this is what you looked for. Maybe it makes sense to implement that structure as an independent layer in mlpack.

I am still stuck at this and don't exactly know what to do, it would be easy if we somehow had a way to define the flow for the network but that is not how it is designed and the main problem here is the downsampling block, i can put all of the things inside a residual block and say it saves the input of the first layer into a temp variable and then tries to add it to the last layer in the residual block but when it does that it finds out that the shapes don't match and i don't see how i can use AddMerge to achieve the same flow. but do let me know if you see some other way around it. I have been thinking around it way too long.

kartikdutt18:
I will try to think of a solution for this and get back to you.

zoq:
Just to make sure, I get what you are trying to do, in some cases the output of the sequential part doesn't match with the input so if you add the skip connection you have to add another layer to convert the input?

Something like:

Unbenannte Zeichnung

Aakash-kaushik:
> Just to make sure, I get what you are trying to do, in some cases the output of the sequential part doesn't match with the input so if you add the skip connection you have to add another layer to convert the input?
> 
> Something like:
> 
> Unbenannte Zeichnung

Yes i believe this is exactly what i am trying to do. 
this was a great diagram for that. Thank you so much.

Aakash-kaushik:
> Just to make sure, I get what you are trying to do, in some cases the output of the sequential part doesn't match with the input so if you add the skip connection you have to add another layer to convert the input?
> 
> Something like:
> 
> Unbenannte Zeichnung

Also @zoq if you have time, can you explain how the diagram you made would work in code and also explain a little bit about the code that you wrote with addmerge and residual layer together.

zoq:
Sure, if you have a specific example you like me to show here, let me know if not I just keep it general.

Aakash-kaushik:
> Sure, if you have a specific example you like me to show here, let me know if not I just keep it general.

Yup let me reference a code block from torchvision that i would like to have, that should give a better idea. 
https://github.com/pytorch/vision/blob/master/torchvision/models/resnet.py#L59-L83

Aakash-kaushik:
> > Sure, if you have a specific example you like me to show here, let me know if not I just keep it general.
> 
> Yup let me reference a code block from torchvision that i would like to have, that should give a better idea.
> https://github.com/pytorch/vision/blob/master/torchvision/models/resnet.py#L59-L83

btw this was just to give you an idea on what i was thinking upon any example that is similar or close should work for me.

kartikdutt18:
```AddMerge<> resblock(false, false);

Sequential<>* sequential = new Sequential<>(true);

Convolution<> convA = new Convolution<>(...);
Convolution<> convA = new Convolution<>(...);

sequential->Add(convA);
sequential->Add(convA);

Sequential<>* downsample_layer = this->Downsample(...)

resblock.Add(sequential);
resblock.Add(downsample_layer);```

**Comment by Aakash-kaushik** [comment]
Hey @zoq don't post the code example anymore, i had a meet with kartik and we figured out the residual block so i can implement it now.

**Comment by zoq** [comment]
Sounds great.

**Comment by zoq** [comment]
Thanks for adding the code, this makes it super easy to understand in what direction this is moving forward.

**Comment by Aakash-kaushik** [comment]
> Thanks for adding the code, this makes it super easy to understand in what direction this is moving forward.

Yes I will keep the PR open as our discussion forum and when I am done I can create separate PR that reviews the code and functionality this is more to just have a daily update on what I worked on that day.

**Comment by Aakash-kaushik** [comment]
Just pushed bottleneck block and wanted to confirm two things:
1. Should the addMerge block should be left with it's default setting for our purposes?
2. the identity layer only acts as a layer that that passes the input and gradients without any modifications, right ? I mean it is evident but i still wanted to confirm it once.

**Comment by zoq** [comment]
> Just pushed bottleneck block and wanted to confirm two things:
> 
> 1. Should the addMerge block should be left with it's default setting for our purposes?

With default parameter you mean `model` and `run`? If that is the case, you have to set `model = true` and `run = true`, you only have to set `model=false` if the layers you passed are initialized as part of another layer, e.g. if you pass the same layer to another `AddMerge` layer, in this case you don't want to initialize the layer twice.

> 2. the identity layer only acts as a layer that that passes the input and gradients without any modifications, right ? I mean it is evident but i still wanted to confirm it once.

That is correct.

File: models/resnet/resnet.hpp [hunk 1]
```diff
@@ -0,0 +1,181 @@
...
+        strideWidth, strideHeight, padW, padH, inputWidth, inputHeight,
+        paddingType);
+    downSample->Add(BatchNorm<>(outSize));
+  }
+
+  void BasicBlock()
+  {
+    ann::Sequential<>* basicBlock = new Sequential<>;
+    // Should the addMerge block be left to it's default settings which is
+    // model = false, run = true ?
```
zoq, line 134: See comment below.

models/resnet/resnet_impl.hpp
```diff
@@ -0,0 +1,83 @@
...
+    inputChannel(std::get<0>(inputShape)),
+    inputWidth(std::get<1>(inputShape)),
+    inputHeight(std::get<2>(inputShape)),
+    numClasses(numClasses)
+{
+
+  resNet.Add(ann::Convolution<>(3, 64, 7, 7, 2, 2, 3, 3, inputWidth, inputHeight));
+  resNet.Add(ann::batchNorm<>(64));
+  resNet.Add(ann::ReLULayer<>);
+  // how to add a padding of 1 for the maxpool operation ?
```
kartikdutt18, line 69: mlpack has Padding layer, is there an issue with using it?

Aakash-kaushik: I assumed that it was a utility function and didn't see that it was a layer, if there is i believe that should work.

zoq: Right, you can use https://github.com/mlpack/mlpack/blob/master/src/mlpack/methods/ann/layer/padding.hpp.

Aakash-kaushik:
> > Just pushed bottleneck block and wanted to confirm two things:
> > 
> > 1. Should the addMerge block should be left with it's default setting for our purposes?
> 
> With default parameter you mean `model` and `run`? If that is the case, you have to set `model = true` and `run = true`, you only have to set `model=false` if the layers you passed are initialized as part of another layer, e.g. if you pass the same layer to another `AddMerge` layer, in this case you don't want to initialize the layer twice.
> 
> > 1. the identity layer only acts as a layer that that passes the input and gradients without any modifications, right ? I mean it is evident but i still wanted to confirm it once.
> 
> That is correct.

thanks

models/resnet/resnet_impl.hpp
Aakash-kaushik: Thanks, i got it. and i have also pushed some code but i doubt if the padding in conv layer which is `padW` and `padH` if i specify both of them to be 1 and 1 will that mean something like this 
original matrix : `4*4`
new matrix with `padW = 1` and `padH = 1` : `6*6` 
and if not does the `padding` param equals 1 for [pytroch's conv](https://pytorch.org/docs/stable/generated/torch.nn.Conv2d.html) layer equal to `padW` = 1 and `padH` = 1 in our case ?

models/resnet/resnet_impl.hpp
```diff
@@ -0,0 +1,103 @@
...
+  {
+    size_t numBlockArray = [2 ,2, 2, 2]  
+  }
+
+  MakeLayer("basicblock", 64, numBlockArray[0]);
+  MakeLayer("basicblock", 128, numBlockArray[1]);
+  MakeLayer("basicblock", 256, numBlockArray[2]);
+  MakeLayer("basicblock", 512, numBlockArray[3]);
+
+  // What would be the Pytroch equivalent of nn.AdaptiveAvgPool2d((1, 1))
```
Aakash-kaushik, line 88: And also a doubt in this.

Aakash-kaushik: Hey @zoq, @kartikdutt18 i had these two doubts.

zoq: I sorry I forget to provide an answer for that one, there is no `AdaptivePooling` in mlpack, but as long as the input size doesn't change you can use `AveragePooling`, but we have to make sure we don't pass different image sizes to the model.

kartikdutt18: I think Adaptive Pooling is Adaptive Max Pooling or Adaptive Mean Pooling. Link : https://github.com/mlpack/mlpack/blob/master/src/mlpack/methods/ann/layer/adaptive_max_pooling_impl.hpp

Aakash-kaushik: Hey @zoq as @kartikdutt18 pointed out we do have adaptive max and mean pools and i have used that one only but for now i have seg faults in my model and no good way to debug it gdb throws out bunch of errors and i am not too experienced with it, so how should i go forward now ? I was also thinking to take a look at implementing something in the ffn class in mlpack to print model summary.

kartikdutt18: Try building model block by block and call forward propagation at after each block to find out at which block the model segfaults. Use that info to determine cause.

zoq: > I think Adaptive Pooling is Adaptive Max Pooling or Adaptive Mean Pooling. Link : https://github.com/mlpack/mlpack/blob/master/src/mlpack/methods/ann/layer/adaptive_max_pooling_impl.hpp

Ohh, I missed that one, `AdaptiveMeanPooling` would be the equivalent then.

Aakash-kaushik: > Try building model block by block and call forward propagation, after adding each block to find out at which block the model segfaults. Use that info to determine cause.

Wouldn't building a model summary like thing be more helpful, that would give me each layer and their params. and would kill 2 things with one. Also i don't feel like that maybe too much work for just debugging and would at the end be deleted.

kartikdutt18: Try one thing first, add cout statements in blocks and print block name and input and output sizes. Should be sufficient to debug.

Aakash-kaushik: Yup that should be better. thanks.

Aakash-kaushik: Prints out a simple layer with in and out size and that was incorrect made that work but still have a seg fault and now i believe that it is mostly due to my lack of undertanding of how padding or certain params work in mlpack and that is messing up with the dimensions. btw also slight congo we were able to replicate the resnet18 architecture and so the other ones should be easy once i can figure the seg fault.

Aakash-kaushik:
Work left: 
1. Figure seg fault (basically make resnet 18 work)
2. Write bottlenect block (skipped it because wasn't needed for resnet 18 or 34 and i want to make these work first)
3. just write some if else for all resnets, that is basically the number of blocks and nothing else 
4. Port weights from pytorch to mlpack. 

With these 4 things the ResNet module should be done. 
Just writing these out because for some reason i tend to miss things in the meet.

Aakash-kaushik:
For some reason the input width and height keep on increasing while i go down in the architecture. adding the output for this and also updating the code. 

the last two figure in layers are the inputWidth and inputHeight respectively. 
```
Convolution: 3 64 7 7 2 2 3 3 224 224
BatchNorm: 64
Relu
Padding: 1,1,1,1 114 114
MaxPool: 3,3,2,2 56 56
Convolution: 64 64 1 1 1 1 1 1 56 56
BatchNorm: 64
Relu
Convolution: 64 64 3 3 1 1 1 1 58 58
BatchNorm: 64
IdentityLayer
Relu
Convolution: 64 64 1 1 1 1 1 1 58 58
BatchNorm: 64
Relu
Convolution: 64 64 3 3 1 1 1 1 60 60
BatchNorm: 64
IdentityLayer
Relu
Convolution: 64 128 1 1 1 1 1 1 60 60
BatchNorm: 128
Relu
Convolution: 128 128 3 3 1 1 1 1 62 62
BatchNorm: 128
DownSample below
Convolution: 64 128 1 1 1 1 0 0 62 62
BatchNorm: 128
Relu
Convolution: 128 128 1 1 1 1 1 1 62 62
BatchNorm: 128
Relu
Convolution: 128 128 3 3 1 1 1 1 64 64
BatchNorm: 128
IdentityLayer
Relu
Convolution: 128 256 1 1 1 1 1 1 64 64
BatchNorm: 256
Relu
Convolution: 256 256 3 3 1 1 1 1 66 66
BatchNorm: 256
DownSample below
Convolution: 128 256 1 1 1 1 0 0 66 66
BatchNorm: 256
Relu
Convolution: 256 256 1 1 1 1 1 1 66 66
BatchNorm: 256
Relu
Convolution: 256 256 3 3 1 1 1 1 68 68
BatchNorm: 256
IdentityLayer
Relu
Convolution: 256 512 1 1 1 1 1 1 68 68
BatchNorm: 512
Relu
Convolution: 512 512 3 3 1 1 1 1 70 70
BatchNorm: 512
DownSample below
Convolution: 256 512 1 1 1 1 0 0 70 70
BatchNorm: 512
Relu
Convolution: 512 512 1 1 1 1 1 1 70 70
BatchNorm: 512
Relu
Convolution: 512 512 3 3 1 1 1 1 72 72
BatchNorm: 512
IdentityLayer
Relu
AdaptiveMeanPooling: 1,1
Linear: 512 1000
Segmentation fault (core dumped)
```

kartikdutt18:
Have you compared it with PyTorch?

Aakash-kaushik:
> Have you compared it with PyTorch?

I did, btw Pytorch doesn't shows the input dimensions so did that with a tf implementation with same input size and as it is obvious the dimensions which are width and height should decrease while the channels increases as we go.

Aakash-kaushik:
> For some reason the input width and height keep on increasing while i go down in the architecture. adding the output for this and also updating the code.
> 
> the last two figure in layers are the inputWidth and inputHeight respectively.
> 
> ```
> Convolution: 3 64 7 7 2 2 3 3 224 224
> BatchNorm: 64
> Relu
> Padding: 1,1,1,1 114 114
> MaxPool: 3,3,2,2 56 56
> Convolution: 64 64 1 1 1 1 1 1 56 56
> BatchNorm: 64
> Relu
> Convolution: 64 64 3 3 1 1 1 1 58 58
> BatchNorm: 64
> IdentityLayer
> Relu
> Convolution: 64 64 1 1 1 1 1 1 58 58
> BatchNorm: 64
> Relu
> Convolution: 64 64 3 3 1 1 1 1 60 60
> BatchNorm: 64
> IdentityLayer
> Relu
> Convolution: 64 128 1 1 1 1 1 1 60 60
> BatchNorm: 128
> Relu
> Convolution: 128 128 3 3 1 1 1 1 62 62
> BatchNorm: 128
> DownSample below
> Convolution: 64 128 1 1 1 1 0 0 62 62
> BatchNorm: 128
> Relu
> Convolution: 128 128 1 1 1 1 1 1 62 62
> BatchNorm: 128
> Relu
> Convolution: 128 128 3 3 1 1 1 1 64 64
> BatchNorm: 128
> IdentityLayer
> Relu
> Convolution: 128 256 1 1 1 1 1 1 64 64
> BatchNorm: 256
> Relu
> Convolution: 256 256 3 3 1 1 1 1 66 66
> BatchNorm: 256
> DownSample below
> Convolution: 128 256 1 1 1 1 0 0 66 66
> BatchNorm: 256
> Relu
> Convolution: 256 256 1 1 1 1 1 1 66 66
> BatchNorm: 256
> Relu
> Convolution: 256 256 3 3 1 1 1 1 68 68
> BatchNorm: 256
> IdentityLayer
> Relu
> Convolution: 256 512 1 1 1 1 1 1 68 68
> BatchNorm: 512
> Relu
> Convolution: 512 512 3 3 1 1 1 1 70 70
> BatchNorm: 512
> DownSample below
> Convolution: 256 512 1 1 1 1 0 0 70 70
> BatchNorm: 512
> Relu
> Convolution: 512 512 1 1 1 1 1 1 70 70
> BatchNorm: 512
> Relu
> Convolution: 512 512 3 3 1 1 1 1 72 72
> BatchNorm: 512
> IdentityLayer
> Relu
> AdaptiveMeanPooling: 1,1
> Linear: 512 1000
> Segmentation fault (core dumped)
> ```

Hey @kartikdutt18, @zoq the part where the `inputWidth` and `outputWidth` is updated inside the conv3x3 and conv1x1 functions that isin't actually updating the variables, do you guys know why would that be ?

zoq:
You mean https://github.com/mlpack/models/pull/61/files#diff-a516cb18be832520af513b17fdd1b96f5e140f661a1f5c34a69f3e4e5a3c19b3R90-R92 is not updating?

Aakash-kaushik:
> You mean https://github.com/mlpack/models/pull/61/files#diff-a516cb18be832520af513b17fdd1b96f5e140f661a1f5c34a69f3e4e5a3c19b3R90-R92 is not updating?

Yes, I can paste a output and you will see that before and after the operation the dimensions are still the same.

Aakash-kaushik:
> > You mean https://github.com/mlpack/models/pull/61/files#diff-a516cb18be832520af513b17fdd1b96f5e140f661a1f5c34a69f3e4e5a3c19b3R90-R92 is not updating?
> 
> Yes, I can paste a output and you will see that before and after the operation the dimensions are still the same.
> I am on windows right now, but will paste an output in sometime.

```
Convolution: 3 64 7 7 2 2 3 3 224 224
BatchNorm: 64
Relu
Padding: 1,1,1,1 114 114
MaxPool: 3,3,2,2 56 56
Convolution: 64 64 3 3 1 1 1 1 56 56
56 56
BatchNorm: 64
Relu
Convolution: 64 64 3 3 1 1 1 1 56 56
56 56
BatchNorm: 64
IdentityLayer
Relu
Convolution: 64 64 3 3 1 1 1 1 56 56
56 56
BatchNorm: 64
Relu
Convolution: 64 64 3 3 1 1 1 1 56 56
56 56
BatchNorm: 64
IdentityLayer
Relu
Convolution: 64 128 3 3 1 1 1 1 56 56
56 56
BatchNorm: 128
Relu
Convolution: 128 128 3 3 1 1 1 1 56 56
56 56
BatchNorm: 128
DownSample below
Convolution: 64 128 1 1 1 1 0 0 56 56
56 56
BatchNorm: 128
Relu
Convolution: 128 128 3 3 1 1 1 1 56 56
56 56
BatchNorm: 128
Relu
Convolution: 128 128 3 3 1 1 1 1 56 56
56 56
BatchNorm: 128
IdentityLayer
Relu
Convolution: 128 256 3 3 1 1 1 1 56 56
56 56
BatchNorm: 256
Relu
Convolution: 256 256 3 3 1 1 1 1 56 56
56 56
BatchNorm: 256
DownSample below
Convolution: 128 256 1 1 1 1 0 0 56 56
56 56
BatchNorm: 256
Relu
Convolution: 256 256 3 3 1 1 1 1 56 56
56 56
BatchNorm: 256
Relu
Convolution: 256 256 3 3 1 1 1 1 56 56
56 56
BatchNorm: 256
IdentityLayer
Relu
Convolution: 256 512 3 3 1 1 1 1 56 56
56 56
BatchNorm: 512
Relu
Convolution: 512 512 3 3 1 1 1 1 56 56
56 56
BatchNorm: 512
DownSample below
Convolution: 256 512 1 1 1 1 0 0 56 56
56 56
BatchNorm: 512
Relu
Convolution: 512 512 3 3 1 1 1 1 56 56
56 56
BatchNorm: 512
Relu
Convolution: 512 512 3 3 1 1 1 1 56 56
56 56
BatchNorm: 512
IdentityLayer
Relu
AdaptiveMeanPooling: 1,1
Linear: 512 1000
```

that `56 56` you see after layer is just printed after the updation of inputWidth and inputheight but as you can see it's still the same.

Aakash-kaushik:
> > > You mean https://github.com/mlpack/models/pull/61/files#diff-a516cb18be832520af513b17fdd1b96f5e140f661a1f5c34a69f3e4e5a3c19b3R90-R92 is not updating?
> > 
> > 
> > Yes, I can paste a output and you will see that before and after the operation the dimensions are still the same.
> > I am on windows right now, but will paste an output in sometime.
> 
> ```
> Convolution: 3 64 7 7 2 2 3 3 224 224
> BatchNorm: 64
> Relu
> Padding: 1,1,1,1 114 114
> MaxPool: 3,3,2,2 56 56
> Convolution: 64 64 3 3 1 1 1 1 56 56
> 56 56
> BatchNorm: 64
> Relu
> Convolution: 64 64 3 3 1 1 1 1 56 56
> 56 56
> BatchNorm: 64
> IdentityLayer
> Relu
> Convolution: 64 64 3 3 1 1 1 1 56 56
> 56 56
> BatchNorm: 64
> Relu
> Convolution: 64 64 3 3 1 1 1 1 56 56
> 56 56
> BatchNorm: 64
> IdentityLayer
> Relu
> Convolution: 64 128 3 3 1 1 1 1 56 56
> 56 56
> BatchNorm: 128
> Relu
> Convolution: 128 128 3 3 1 1 1 1 56 56
> 56 56
> BatchNorm: 128
> DownSample below
> Convolution: 64 128 1 1 1 1 0 0 56 56
> 56 56
> BatchNorm: 128
> Relu
> Convolution: 128 128 3 3 1 1 1 1 56 56
> 56 56
> BatchNorm: 128
> Relu
> Convolution: 128 128 3 3 1 1 1 1 56 56
> 56 56
> BatchNorm: 128
> IdentityLayer
> Relu
> Convolution: 128 256 3 3 1 1 1 1 56 56
> 56 56
> BatchNorm: 256
> Relu
> Convolution: 256 256 3 3 1 1 1 1 56 56
> 56 56
> BatchNorm: 256
> DownSample below
> Convolution: 128 256 1 1 1 1 0 0 56 56
> 56 56
> BatchNorm: 256
> Relu
> Convolution: 256 256 3 3 1 1 1 1 56 56
> 56 56
> BatchNorm: 256
> Relu
> Convolution: 256 256 3 3 1 1 1 1 56 56
> 56 56
> BatchNorm: 256
> IdentityLayer
> Relu
> Convolution: 256 512 3 3 1 1 1 1 56 56
> 56 56
> BatchNorm: 512
> Relu
> Convolution: 512 512 3 3 1 1 1 1 56 56
> 56 56
> BatchNorm: 512
> DownSample below
> Convolution: 256 512 1 1 1 1 0 0 56 56
> 56 56
> BatchNorm: 512
> Relu
> Convolution: 512 512 3 3 1 1 1 1 56 56
> 56 56
> BatchNorm: 512
> Relu
> Convolution: 512 512 3 3 1 1 1 1 56 56
> 56 56
> BatchNorm: 512
> IdentityLayer
> Relu
> AdaptiveMeanPooling: 1,1
> Linear: 512 1000
> ```
> 
> that `56 56` you see after layer is just printed after the updation of inputWidth and inputheight but as you can see it's still the same.

Seems like i am wrong and it is updating the params but they are just coming up to be 56 again. Will look into it.

zoq:
If you debug something like that, I would recommend to start with something smaller, like in this case it's the first Conv layer, so you don't need the other layers.

Aakash-kaushik:
> If you debug something like that, I would recommend to start with something smaller, like in this case it's the first Conv layer, so you don't need the other layers.

Hi, thanks for this the reason why i wasn't doing this was because i thought the predict function or something at the end expects a certain way the matrix should be but i see we don't use a two dimensional array similar to other libraries to represent width and dimensions. Because they are 4 dimensional (channel, width, height, batch) but i read further mlpack code and saw that we have (channel*width*height, batch)

Aakash-kaushik:
> If you debug something like that, I would recommend to start with something smaller, like in this case it's the first Conv layer, so you don't need the other layers.

So I did this and as soon as I wrap even a single layer into the sequential layer and add that to the network I get a segmentation error and I have taken care of the point where it should be preceded with an identity layer if it is used as the first layer which it isn't and the second point where it needs at least two layers to call its gradient function I have added two-layer which work when independently added to the network but not when added in the sequential block and also I don't think the predict function calls the gradient function.

Aakash-kaushik:
The last commit id for my mlpack version: 58869b57c50926c4e9682c35c6c6efa8ea31d604
https://github.com/mlpack/mlpack/commit/58869b57c50926c4e9682c35c6c6efa8ea31d604

I believe it's the latest one that was merged 3 days ago.

zoq:
> The last commit id for my mlpack version: 58869b57c50926c4e9682c35c6c6efa8ea31d604
> [mlpack/mlpack@58869b5](https://github.com/mlpack/mlpack/commit/58869b57c50926c4e9682c35c6c6efa8ea31d604)
> 
> I believe it's the latest one that was merged 3 days ago.

Thanks with that I'm able to reproduce the issue on my system.

Aakash-kaushik:
> > The last commit id for my mlpack version: 58869b57c50926c4e9682c35c6c6efa8ea31d604
> > [mlpack/mlpack@58869b5](https://github.com/mlpack/mlpack/commit/58869b57c50926c4e9682c35c6c6efa8ea31d604)
> > I believe it's the latest one that was merged 3 days ago.
> 
> Thanks with that I'm able to reproduce the issue on my system.

That's great, I did try a couple more things yesterday but to no avail.

Some of them were: 
1. Adding a identity layer at the start.
2. Wrapping the first layers in a sequential block So the network starts with a sequential block.

Things i will try:
1. Just going to run a simple sequential block now in a ffn class.

Aakash-kaushik:
Hey @zoq, @kartikdutt18 i figured it out. 
I feel happy and dumb at the same time. 
So in the get model function it was just returning the ffn object rather than the reference of it.

Aakash-kaushik:
Had this 
```c++
ann::FFN<OutputLayerType, InitializationRuleType> GetModel()
      { return resNet; }
```

When it should have been this: 
```c++
ann::FFN<OutputLayerType, InitializationRuleType>& GetModel()
      { return resNet; }
```

Aakash-kaushik:
So i am using this padding layer this way 
```c++
resNet.Add(new ann::Padding<>(1, 1, 1, 1));
```
and i checked the output shape before the padding layer and after the padding layer they are both the same. the way i did this was to just comment  out the padding layer from resnet so it was never added and see the output shape and second by just adding it so i think i might not be using it in the right way.

kartikdutt18:
Was `includeTop` set to `True`?

Aakash-kaushik:
> Was `includeTop` set to `True`?

Hey, as you can see the include part comes later in the code, I just updated the code for resnet_impl so you can see that, i am referencing this line: https://github.com/Aakash-kaushik/models/blob/resnet/models/resnet/resnet_impl.hpp#L111

zoq:
All the padding layer does is to pad the input and put the original data in the center:

https://github.com/mlpack/mlpack/blob/58869b57c50926c4e9682c35c6c6efa8ea31d604/src/mlpack/methods/ann/layer/padding_impl.hpp#L43-L48

can you confirm that `nRows = input.n_rows;  nCols = input.n_cols;` is what you see as input, from the arguments the expected output should be increased by one after the layer right?

zoq:
> > All the padding layer does is to pad the input and put the original data in the center:
> > https://github.com/mlpack/mlpack/blob/58869b57c50926c4e9682c35c6c6efa8ea31d604/src/mlpack/methods/ann/layer/padding_impl.hpp#L43-L48
> > can you confirm that `nRows = input.n_rows; nCols = input.n_cols;` is what you see as input, from the arguments the expected output should be increased by one after the layer right?
> 
> Hey would it be possible for you to come for a meet today ? It shouldn't take more than 10 mins and is a bit to explain this way

I could hope on a call real quick.

Aakash-kaushik:
> > > All the padding layer does is to pad the input and put the original data in the center:
> > > https://github.com/mlpack/mlpack/blob/58869b57c50926c4e9682c35c6c6efa8ea31d604/src/mlpack/methods/ann/layer/padding_impl.hpp#L43-L48
> > > can you confirm that `nRows = input.n_rows; nCols = input.n_cols;` is what you see as input, from the arguments the expected output should be increased by one after the layer right?
> > 
> > 
> > Hey would it be possible for you to come for a meet today ? It shouldn't take more than 10 mins and is a bit to explain this way
> 
> I could hope on a call real quick.

Would 8:30 pm IST work for you, basically 27 mins from now ?

zoq:
> > > > All the padding layer does is to pad the input and put the original data in the center:
> > > > https://github.com/mlpack/mlpack/blob/58869b57c50926c4e9682c35c6c6efa8ea31d604/src/mlpack/methods/ann/layer/padding_impl.hpp#L43-L48
> > > > can you confirm that `nRows = input.n_rows; nCols = input.n_cols;` is what you see as input, from the arguments the expected output should be increased by one after the layer right?
> > > 
> > > 
> > > Hey would it be possible for you to come for a meet today ? It shouldn't take more than 10 mins and is a bit to explain this way
> > 
> > 
> > I could hope on a call real quick.
> 
> Would 8:30 pm IST work for you, basically 27 mins from now ?

That works for me.

Aakash-kaushik:
> > > > > All the padding layer does is to pad the input and put the original data in the center:
> > > > > https://github.com/mlpack/mlpack/blob/58869b57c50926c4e9682c35c6c6efa8ea31d604/src/mlpack/methods/ann/layer/padding_impl.hpp#L43-L48
> > > > > can you confirm that `nRows = input.n_rows; nCols = input.n_cols;` is what you see as input, from the arguments the expected output should be increased by one after the layer right?
> > > > 
> > > > 
> > > > Hey would it be possible for you to come for a meet today ? It shouldn't take more than 10 mins and is a bit to explain this way
> > > 
> > > 
> > > I could hope on a call real quick.
> > 
> > 
> > Would 8:30 pm IST work for you, basically 27 mins from now ?
> 
> That works for me.

Btw can't invite you because I can't see your email id anymore, so i think we can meet on the mlpack zoom room.

zoq:
> > > > > > All the padding layer does is to pad the input and put the original data in the center:
> > > > > > https://github.com/mlpack/mlpack/blob/58869b57c50926c4e9682c35c6c6efa8ea31d604/src/mlpack/methods/ann/layer/padding_impl.hpp#L43-L48
> > > > > > can you confirm that `nRows = input.n_rows; nCols = input.n_cols;` is what you see as input, from the arguments the expected output should be increased by one after the layer right?
> > > > > 
> > > > > 
> > > > > Hey would it be possible for you to come for a meet today ? It shouldn't take more than 10 mins and is a bit to explain this way
> > > > 
> > > > 
> > > > I could hope on a call real quick.
> > > 
> > > 
> > > Would 8:30 pm IST work for you, basically 27 mins from now ?
> > 
> > 
> > That works for me.
> 
> Btw can't invite you because I can't see your email id anymore, so i think we can meet on the mlpack zoom room.

The zoom rooms works.

Aakash-kaushik:
So i found out about the dim mismatch error and i think that is happening because the `AddMerge` layer instead of taking the outptut of the previous layer and sending it to both the `downSample` block and the first `Sequential` block, it is taking the output of it's first layer and sending that to both of them that i what i could get based on what i can see but i don't think that is possible so if you guys are free sometime today we can talk over a meet or we can conitnue here too. because tomorrow is weekend and no taunt to @zoq but i get it, and i am also trying to enjoy over the weekends :rocket:

Aakash-kaushik:
Also i have pushed the code to the latest stage so you can see that too.

kartikdutt18:
Could you share the dim mismatch in a bit more detail here.

Aakash-kaushik:
So this is the output that we get when we have a single `basicBlock`
```
Convolution: 3 64 7 7 2 2 3 3 112 112
BatchNorm: 64
Relu
Padding: 1,1,1,1 114 114
MaxPool: 3,3,2,2 56 56
Convolution: 64 64 3 3 1 1 1 1 56 56
BatchNorm: 64
Relu
Convolution: 64 64 3 3 1 1 1 1 56 56
BatchNorm: 64
IdentityLayer
Relu
Convolution: 64 64 3 3 1 1 1 1 56 56
BatchNorm: 64
Relu
Convolution: 64 64 3 3 1 1 1 1 56 56
BatchNorm: 64
IdentityLayer
Relu
Rows: 200704
Cols: 1
```

The Rows output just comes from multiplying: `56 * 56 * 64 = 200704`

Then below we can see the output that comes out when i add a second `basicBlock`

```
Convolution: 3 64 7 7 2 2 3 3 112 112
BatchNorm: 64
Relu
Padding: 1,1,1,1 114 114
MaxPool: 3,3,2,2 56 56
Convolution: 64 64 3 3 1 1 1 1 56 56
BatchNorm: 64
Relu
Convolution: 64 64 3 3 1 1 1 1 56 56
BatchNorm: 64
IdentityLayer
Relu
Convolution: 64 64 3 3 1 1 1 1 56 56
BatchNorm: 64
Relu
Convolution: 64 64 3 3 1 1 1 1 56 56
BatchNorm: 64
IdentityLayer
Relu
new layer
Convolution: 64 128 3 3 2 2 1 1 28 28
BatchNorm: 128
Relu
Convolution: 128 128 3 3 1 1 1 1 28 28
BatchNorm: 128
DownSample below
Convolution: 64 128 1 1 2 2 0 0 28 28
BatchNorm: 128
Relu
Convolution: 128 128 3 3 1 1 1 1 28 28
BatchNorm: 128
Relu
Convolution: 128 128 3 3 1 1 1 1 28 28
BatchNorm: 128
IdentityLayer
Relu

error: addition: incompatible matrix dimensions: 100352x1 and 25088x1
terminate called after throwing an instance of 'std::logic_error'
  what():  addition: incompatible matrix dimensions: 100352x1 and 25088x1
Aborted (core dumped)
```
Now we see the mismatch error a fig of `100252` comes from multiplying `28 * 28 *128` and if we pass this though the dowsample block than we get `25088` but we don't want to pass the output of the first layer of the second `basicBlock`. what we want to do is pass the `200704` here so it then becomes `100252` after going into the `downSample` block and adds successfully.

Aakash-kaushik:
Let me know if this explains it otherwise a quick call or whatever you prefer works too.

kartikdutt18:
Lets do a quick call on mlpack room in 5 minutes ?

Aakash-kaushik:
> Lets do a quick call on zoom (mlpack room) in 5 minutes ?

Sure, will be there.

Aakash-kaushik:
@kartikdutt18 I am already there.

kartikdutt18:
Joining

zoq:
> So i found out about the dim mismatch error and i think that is happening because the `AddMerge` layer instead of taking the outptut of the previous layer and sending it to both the `downSample` block and the first `Sequential` block, it is taking the output of it's first layer and sending that to both of them that i what i could get based on what i can see but i don't think that is possible so if you guys are free sometime today we can talk over a meet or we can conitnue here too. because tomorrow is weekend and no taunt to @zoq but i get it, and i am also trying to enjoy over the weekends

Let me know if you figured it out.

Aakash-kaushik:
> > So i found out about the dim mismatch error and i think that is happening because the `AddMerge` layer instead of taking the outptut of the previous layer and sending it to both the `downSample` block and the first `Sequential` block, it is taking the output of it's first layer and sending that to both of them that i what i could get based on what i can see but i don't think that is possible so if you guys are free sometime today we can talk over a meet or we can conitnue here too. because tomorrow is weekend and no taunt to @zoq but i get it, and i am also trying to enjoy over the weekends
> 
> Let me know if you figured it out.

Hey, Kartik proposed a solution, I am trying it right now, will let you know in sometime if it was resolved.
Mostly it should be.

zoq:
Okay nice, if not let me know and I'll check if I can find a solution as well.

Aakash-kaushik:
> Okay nice, if not let me know and I'll check if I can find a solution as well.

I guess that worked but we have another thing
```
Convolution: 3 64 7 7 2 2 3 3 112 112
BatchNorm: 64
Relu
Padding: 1,1,1,1 114 114
MaxPool: 3,3,2,2 56 56
Convolution: 64 64 3 3 1 1 1 1 56 56
BatchNorm: 64
Relu
Convolution: 64 64 3 3 1 1 1 1 56 56
BatchNorm: 64
IdentityLayer
Relu
Convolution: 64 64 3 3 1 1 1 1 56 56
BatchNorm: 64
Relu
Convolution: 64 64 3 3 1 1 1 1 56 56
BatchNorm: 64
IdentityLayer
Relu
new layer
Convolution: 64 128 3 3 2 2 1 1 28 28
BatchNorm: 128
Relu
Convolution: 128 128 3 3 1 1 1 1 28 28
BatchNorm: 128
DownSample below
Convolution: 64 128 1 1 2 2 0 0 56 56
BatchNorm: 128
Relu

error: addition: incompatible matrix dimensions: 100352x1 and 200704x1
terminate called after throwing an instance of 'std::logic_error'
  what():  addition: incompatible matrix dimensions: 100352x1 and 200704x1
Aborted (core dumped)
```
anything passed to the downsample should be reduced now but it isin't doing that.

Aakash-kaushik:
Hey @kartikdutt18, @zoq  I have pushed the code too can you take another look ?

kartikdutt18:
I'm out rn, will take a look around 2030 IST.

Aakash-kaushik:
> I'm out rn, will take a look around 2030 IST.
Thanks.

models/resnet/resnet.hpp
```diff
@@ -0,0 +1,310 @@
...
+
+      std::cout<<"Convolution: "<<inSize<<" "<<outSize<<" "<<kernelWidth<<" "<<
+          kernelHeight<<" "<<strideWidth<<" "<<strideHeight<<" "<<padW<<" "<<
+         padH<<" "<<downSampleInputWidth<<" "<<downSampleInputHeight<<std::endl;
+    }
+    else
+    {
+      baseLayer->Add(new ann::Convolution<>(inSize, outSize, kernelWidth,
+          kernelHeight, strideWidth, strideHeight, padW, padH, inputWidth,
+          inputHeight));
```
kartikdutt18, line 125: Try this:
```suggestion
          kernelHeight, strideWidth, strideHeight, padW, padH, downSampleInputWidth,
          downSampleInputHeight));
```

Aakash-kaushik (commented):
ignore this, this was some error as github was forcing me to leave a review.

models/resnet/resnet.hpp
```diff
@@ -0,0 +1,310 @@
...
+                  const size_t strideWidth = 2,
+                  const size_t strideHeight = 2,
+                  const size_t padW = 0,
+                  const size_t padH = 0)
+  {
+    ConvolutionBlock1x1(downSample, inSize, outSize, downSampleInputWidth,
+        downSampleInputHeight, strideWidth, strideHeight, kernelWidth,
+        kernelHeight, padW, padH, true);
+
+    downSample->Add(new ann::BatchNorm<>(outSize));
```
kartikdutt18, line 154: Should batchnorm be added as third connection or as part of downsample. Wrap it as sequential and insert it into downsample layer.

Aakash-kaushik: I just pushed it, can you verify if this is what you meant ?

Aakash-kaushik: It worked !!! :rocket:

kartikdutt18: Cool

Aakash-kaushik: Thank you so much for spotting it.

Aakash-kaushik:
Hey, so all the architectures are ready but the thing that pulls me back a bit is a resnet152 with `224 * 224 * 3` input dim takes up 7.1 GB of RAM, isin't that too much, I am happy i had a swap space otherwise it wouldn't even run, btw this is runtime memory not compile time.

Aakash-kaushik:
Also @kartikdutt18 it would be great if you can walk me through the weight converter because that part is a bit tougher than what thought, i assumed that i would just need to supply the model object and it would save them as a file which i could further edit.

kartikdutt18:
Let's first clean up this PR so it can be reviewed. This included adding comments, making the style check pass and squashing commits.

Aakash-kaushik:
> Let's first clean up this PR so it can be reviewed. This included adding comments, making the style check pass and squashing commits.

btw should i keep the output that is printed now? maybe as mlpack log or something that can be enabled or something like that ?

kartikdutt18:
You can use logs similar to Darknet. I think they are more clear / easier to understand.

Aakash-kaushik:
> You can use logs similar to Darknet. I think they are more clear / easier to understand.

Great, shall do that.

Aakash-kaushik:
btw how do we see the output of mlpack::log::info ? i haven't used it before.

Aakash-kaushik:
I haven't added the pretrained part of code because i didn't really had weights but that can be easily added once we have the code so i don't think much to worry about that.

Aakash-kaushik:
Hey @kartikdutt18, @zoq I have created #63, It wouild be great if we can review the PR over there.

PR #67, opened 2021-06-29
Deprecation of ubuntu 16.04 in pipeline 

Aakash-kaushik:
image
The support will go out in September and azure only uses the LTS versions of ubuntu and tags them with `ubuntu-latest` so using this should fix us and wouldn't need any change or a long time but I can also see the benefit of using `ubuntu-20.04` where it doesn't changes and we can have a testing env we know.   

Let me know your opinions.

### DIFF
--- a/.ci/ci.yaml
+++ b/.ci/ci.yaml
@@ -28,7 +28,7 @@ jobs:
 - job: Linux
   timeoutInMinutes: 360
   pool:
-    vmImage: ubuntu-16.04
+    vmImage: ubuntu-latest
 
   strategy:
     matrix:


.ci/ci.yaml
```diff
@@ -28,7 +28,7 @@ jobs:
 - job: Linux
   timeoutInMinutes: 360
   pool:
-    vmImage: ubuntu-16.04
+    vmImage: ubuntu-latest
```
zoq, line 31: Using `ubuntu-latest` sounds reasonable to me.

Aakash-kaushik: Nice, also this is the same for mlpack too, so should i open a PR there too ?

zoq: For mlpack we wanted to make sure we run tests against the oldest version, so we will wait until September 20.

Aakash-kaushik: Sure, maybe we can leave it that way.

Aakash-kaushik:
All the checks do pass so maybe we can merge this ?

zoq (approved):
Looks great to me, thanks for the update.

zoq:
Thanks again for the patch!

PR #63, opened 2021-06-19
Addition of resnet module. 

Aakash-kaushik:
Extension of #61 and a PR for review of the Additon of resnet module.

PR #69, opened 2021-07-02
Use catch2 instead of ctest

Aakash-kaushik:
ctest reports tests to pass even if they are failed, this is done just so we know if some tests fail. 

reference comment: https://github.com/mlpack/models/pull/63#issuecomment-870899105

### DIFF
--- a/.ci/ci.yaml
+++ b/.ci/ci.yaml
@@ -18,7 +18,7 @@ jobs:
   strategy:
     matrix:
       Plain:
-        CMakeArgs-mlpack: '-DDEBUG=ON -DPROFILE=OFF -DBUILD_PYTHON_BINDINGS=OFF -DBUILD_TESTS=OFF'
+        CMakeArgs-mlpack: '-DDEBUG=OFF -DPROFILE=OFF -DBUILD_PYTHON_BINDINGS=OFF -DBUILD_TESTS=OFF'
         CMakeArgs-models: '-DDEBUG=ON'
 
   steps:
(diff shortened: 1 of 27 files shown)


Aakash-kaushik:
How does this even happen, please ignore this PR for now, i didn't mean to include all the resnet changes.

birm:
Do you need guidance on fixing your branch?

Aakash-kaushik:
> Do you need guidance on fixing your branch?

I would highly appreciate it.

zoq:
Looks like you created a branch from the resnet branch and not from the master branch?

.ci/linux-steps.yaml
```diff
@@ -49,7 +49,7 @@ steps:
   displayName: 'Build models'
 
 # Run CTests.
-- script: cd build/tests/ && sudo CTEST_OUTPUT_ON_FAILURE=1 ctest -T Test .
+- script: cd build/ && ./models_test -x 50
```
zoq, line 52: I think it's `bin/models_test`, you can add a `cd build/ && ls && bin/models_test -x` to list the files in the directory.

Aakash-kaushik:
> Looks like you created a branch from the resnet branch and not from the master branch?

This was a bit more messier than that, if it's fine i would open another PR by tomorrow or just fix this one up, i mean just opening another one would be pretty easy to me

.ci/linux-steps.yaml
Aakash-kaushik: Yup yup it is that, and i did that with this commit: https://github.com/mlpack/models/pull/69/commits/ea0f8165f49375225f0c4263fdecdbc8713309e2 but again i messed this up a bit so tomorrow morning would be a better option.

zoq:
Sure, whatever is easier for you.

PR #70, opened 2021-07-03
Use catch2 instead of ctest

Aakash-kaushik:
ctest reports tests to pass even if they are failed, this is done just so we know if some tests fail. 

reference comment: https://github.com/mlpack/models/pull/63#issuecomment-870899105

reverting back on #69 closed it automatically so creating this PR.

### DIFF
--- a/.ci/linux-steps.yaml
+++ b/.ci/linux-steps.yaml
@@ -49,7 +49,7 @@ steps:
   displayName: 'Build models'
 
 # Run CTests.
-- script: cd build/tests/ && sudo CTEST_OUTPUT_ON_FAILURE=1 ctest -T Test .
+- script: cd build/bin/ && ./models_test -x 50
   displayName: 'Run tests via ctest'
 
 # Publish test results to Azure Pipelines
--- a/.ci/windows-steps.yaml
+++ b/.ci/windows-steps.yaml
@@ -138,6 +138,8 @@ steps:
     dir
     cd bin/
     dir
+    cd Release/
+    dir
     ./models_test.exe -x 50
   displayName: 'Run tests via ctest'
 
--- a/.ci/windows-steps.yaml
+++ b/.ci/windows-steps.yaml
@@ -134,13 +134,8 @@ steps:
 
 # Run tests via ctest.
 - bash: |
-    cd build/
-    dir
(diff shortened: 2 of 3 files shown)


Aakash-kaushik:
the `dir` command in windows is just for debugging and will be removed before the final merge.

.ci/windows-steps.yaml
```diff
@@ -134,8 +134,11 @@ steps:
 
 # Run tests via ctest.
 - bash: |
-    cd build/tests
-    CTEST_OUTPUT_ON_FAILURE=1 ctest -T Test -C Release .
+    cd build/
+    dir
+    cd bin/
+    dir
+    ./models_test.exe -x 50
```
zoq, line 141: `models_test.exe` should be in  the `Release` folder.

Aakash-kaushik:
@zoq should be good to merge now.

zoq (approved):
Nice, thanks for looking into this.

.ci/linux-steps.yaml
```diff
@@ -49,7 +49,7 @@ steps:
   displayName: 'Build models'
 
 # Run CTests.
-- script: cd build/tests/ && sudo CTEST_OUTPUT_ON_FAILURE=1 ctest -T Test .
+- script: cd build/bin/ && ./models_test -x 50
```
rcurtin, line 52: Maybe we should leave off the `-x 50`?  That way the log will contain the logs for all failed tests.  (But maybe there is a reason to add it that I overlooked?)

rcurtin (approved):
Nice, thanks for making this change!

.ci/linux-steps.yaml
Aakash-kaushik: So the -x 50 is included because we would want catch2 to continue even if some tests fail. And suppose you have a couple of tests failing therefore 50 is a big enough number to accomodate them all so you can address them at once.

rcurtin: Got it.  Thanks for the clarification!

Aakash-kaushik: No problem 😃

birm (approved):
Great!

PR #75, opened 2022-03-20
Fix documentation

hello-fri-end:
Version of yolo is taken through the constructor and is not a template parameter.


rcurtin (approved):
Thanks, nice catch!

PR #76, opened 2022-07-25
AlexNet Implementation

shubham1206agra:
Implemented AlexNet using the new ann-vtable structure as a Layer for more straightforward serialization and use.


models/alexnet/alexnet.hpp
```diff
@@ -0,0 +1,145 @@
...
+  //! Take ownership of the layers of the given AlexNetType.
+  AlexNetType(AlexNetType&& other);
+  //! Copy the given AlexNetType.
+  AlexNetType& operator=(const AlexNetType& other);
+  //! Take ownership of the given AlexNetType.
+  AlexNetType& operator=(AlexNetType&& other);
+
+  //! Virtual destructor: delete all held layers.
+  virtual ~AlexNetType()
+  { /* Nothing to do here. */ }
```
kartikdutt18, line 73: nit: Can be moved to same line IG

kartikdutt18 (commented):
Can we also please add a test case for this to tests folder

models/alexnet/alexnet.hpp
```diff
@@ -0,0 +1,145 @@
...
+  //! Serialize the AlexNetType.
+  template<typename Archive>
+  void serialize(Archive& ar, const uint32_t /* version */);
+
+ private:
+  void makeModel()
+  {
+    this->template Add<ann::Convolution>(64, 11, 11, 4, 4, 2, 2);
+    this->template Add<ann::ReLU>();
+    this->template Add<ann::MaxPooling>(3, 3, 2, 2);
```
kartikdutt18, line 105: We can create a function that does this, and call that repeatedly like for example, addConvBlock(inputParams) {
  // Add Conv Block, Relu and MaxPooling if needed
}

models/alexnet/alexnet.hpp
```diff
@@ -0,0 +1,120 @@
...
+
+  //! Virtual destructor: delete all held layers.
+  virtual ~AlexNetType()
+  { /* Nothing to do here. */ }
+
+  //! Create a copy of the AlexNetType (this is safe for polymorphic use).
+  AlexNetType* Clone() const { return new AlexNetType(*this); }
+
+  /**
+   * Get the FFN object representing the network.
```
rcurtin, line 78: We should document that the caller is responsible for deleting the returned network.

models/alexnet/alexnet_impl.hpp
```diff
@@ -0,0 +1,126 @@
...
+template<typename MatType>
+AlexNetType<MatType>::AlexNetType(
+    const size_t numClasses,
+    const bool includeTop) :
+    ann::MultiLayer<MatType>(),
+    numClasses(numClasses),
+    includeTop(includeTop)
+{
+  MakeModel();
+}
```
rcurtin, line 38: I think it's possible to use defaults to combine these two constructors, and then if you wanted you could inline `MakeModel()` directly into the constructor for simplicity, since I think it isn't used anywhere else.

models/alexnet/alexnet_impl.hpp
```diff
@@ -0,0 +1,126 @@
...
+  ar(CEREAL_NVP(includeTop));
+}
+
+template<typename MatType>
+void AlexNetType<MatType>::MakeModel()
+{
+  this->template Add<ann::Convolution>(64, 11, 11, 4, 4, 2, 2);
+  this->template Add<ann::ReLU>();
+  this->template Add<ann::MaxPooling>(3, 3, 2, 2);
+  this->template Add<ann::Convolution>(192, 5, 5, 1, 1, 2, 2);
```
rcurtin, line 101: Shouldn't this use `GroupedConvolution` with two groups here?

models/alexnet/alexnet_impl.hpp
```diff
@@ -0,0 +1,126 @@
...
+  this->template Add<ann::ReLU>();
+  this->template Add<ann::MaxPooling>(3, 3, 2, 2);
+  this->template Add<ann::Convolution>(384, 3, 3, 1, 1, 1, 1);
+  this->template Add<ann::ReLU>();
+  this->template Add<ann::Convolution>(256, 3, 3, 1, 1, 1, 1);
+  this->template Add<ann::ReLU>();
+  this->template Add<ann::Convolution>(256, 3, 3, 1, 1, 1, 1);
```
rcurtin, line 108: Are these the right number of filters for each layer?  I am looking at https://proceedings.neurips.cc/paper/2012/file/c399862d3b9d6b76c8436e924a68c45b-Paper.pdf, Figure 2, and it seems off to me.  For instance, I believe the first layer should be a grouped convolution (with 2 groups), with 96 kernels (48 in each group), not 64.  Correct me if I overlooked something, or if I am looking in the wrong place.

rcurtin (commented):
Nice, looking good!  I have some comments here that also are relevant to the VGG PR, and I agree with @kartikdutt18 that we should add some tests here. :+1:

models/alexnet/alexnet_impl.hpp
shubham1206agra: Idk why torchvision has a different architecture
https://github.com/pytorch/vision/blob/main/torchvision/models/alexnet.py
I don't think anybody is using Grouped Conv for AlexNet for some reason.

models/alexnet/alexnet.hpp
```diff
@@ -0,0 +1,120 @@
...
+namespace models {
+
+/**
+ * Definition of a AlexNet CNN.
+ * 
+ * @tparam MatType Matrix representation to accept as input and use for
+ *    computation.
+ */
+template<typename MatType = arma::mat>
+class AlexNetType : public ann::MultiLayer<MatType>
```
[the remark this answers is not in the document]
shubham1206agra: No need for a Gradient check as we have performed gradient checks for every layer beforehand.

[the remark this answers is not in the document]
shubham1206agra: Use here is to use these networks as a layer. So if you just want architecture with no classifier and add something on your own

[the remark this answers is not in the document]
rcurtin: Thanks for the clarification; that makes sense to me.

models/alexnet/alexnet_impl.hpp
rcurtin: Fair enough, if this matches the torchvision reference implementation that's good enough for me. :+1: I guess maybe that was done because I think grouped convolution was only used in AlexNet as a way to distribute across two GPUs.

tests/alexnet_tests.cpp
```diff
@@ -0,0 +1,157 @@
...
+  CheckMatrices(originalOutput, modelOutput, 1e-5);
+  CheckMatrices(originalOutput, xmlOutput, 1e-5);
+  CheckMatrices(originalOutput, jsonOutput, 1e-5);
+  CheckMatrices(originalOutput, binaryOutput, 1e-5);
+}
+
+TEST_CASE("AlexNetSerializationTest", "[AlexnetTests]")
+{
+	models::AlexNet model;
+	ModelSerializationTest(model);
```
rcurtin, line 112: There are some tabbing issues here, can you fix those please? :)

rcurtin (commented):
Everything looks good; if you can handle the couple outstanding comments I think this is ready to merge.

kartikdutt18 (approved):

rcurtin:
Awesome work @shubham1206agra! :+1:

PR #78, opened 2022-07-25
VGG Implementation

shubham1206agra:
Implemented VGG using the new ann-vtable structure as a Layer for more straightforward serialization and use.

### DIFF
--- a/models/vgg/vgg.hpp
+++ b/models/vgg/vgg.hpp
@@ -129,7 +129,7 @@ class VGGType : public ann::MultiLayer<MatType>
       {
         this->template Add<ann::Convolution>(layers[i], 3, 3, 1, 1, 1, 1);
         if (IsBatchNorm)
-          this->template Add<ann::BatchNorm>(layers[i]);
+          this->template Add<ann::BatchNorm>();
         this->template Add<ann::ReLU>();
       }
     }
--- a/models/vgg/vgg.hpp
+++ b/models/vgg/vgg.hpp
@@ -66,7 +66,7 @@ class VGGType : public ann::MultiLayer<MatType>
   VGGType(
     const size_t numClasses,
     const bool includeTop = true);
-  
+
   //! Copy the given VGGType.
   VGGType(const VGGType& other);
   //! Take ownership of the layers of the given VGGType.
@@ -78,9 +78,7 @@ class VGGType : public ann::MultiLayer<MatType>
 
   //! Virtual destructor: delete all held layers.
   virtual ~VGGType()
-  {
-    // Nothing to do here. 
-  }
+  { /* Nothing to do here. */ }
 
   //! Create a copy of the VGGType (this is safe for polymorphic use).
   VGGType* Clone() const { return new VGGType(*this); }
@@ -97,7 +95,7 @@ class VGGType : public ann::MultiLayer<MatType>
   >
   ann::FFN<OutputLayerType, InitializationRuleType, MatType>* GetModel()
   {
-    ann::FFN<OutputLayerType, InitializationRuleType, MatType>* vgg = 
+    ann::FFN<OutputLayerType, InitializationRuleType, MatType>* vgg =
         new ann::FFN<OutputLayerType, InitializationRuleType, MatType>();
     vgg->Add(this);
     return vgg;
@@ -119,7 +117,7 @@ class VGGType : public ann::MultiLayer<MatType>
           0, 512, 512, 512, 512, 0} }
     };
     std::vector<size_t> layers = construct.at(VGGVersion);
-    for (size_t i = 0; i < layers.size(); i++) 
+    for (size_t i = 0; i < layers.size(); i++)
     {
       if (layers[i] == 0)
       {
--- a/models/vgg/vgg.hpp
+++ b/models/vgg/vgg.hpp
@@ -118,7 +118,7 @@ class VGGType : public ann::MultiLayer<MatType>
       { 19, {64, 64, 0, 128, 128, 0, 256, 256, 256, 256, 0, 512, 512, 512, 512,
           0, 512, 512, 512, 512, 0} }
     };
-    std::vector<size_t> layers = construct[VGGVersion];
+    std::vector<size_t> layers = construct.at(VGGVersion);
     for (size_t i = 0; i < layers.size(); i++) 
     {
       if (layers[i] == 0)
--- a/models/vgg/vgg.hpp
+++ b/models/vgg/vgg.hpp
@@ -38,6 +38,12 @@ namespace models {
 /**
  * Definition of a VGG CNN.
  * 
+ * Paper uses different kind of notation for specifying the CNN (Table 1).
+ * A - 11
+ * B - 13
+ * D - 16
+ * E - 19
+ * 
  * @tparam MatType Matrix representation to accept as input and use for
  *    computation.
  * @tparam VGGVersion Version of VGG (Valid Configuration - 11, 13, 16, 19).
@@ -51,17 +57,14 @@ template<
 class VGGType : public ann::MultiLayer<MatType>
 {
  public:
-  //! Create the VGGType layer.
-  VGGType();
-
   /**
-   * VGGType constructor intializes number of classes and weights.
+   * VGGType constructor intializes layer with number of classes given.
    *
-   * @param numClasses Optional number of classes to classify images into,
+   * @param numClasses Number of classes to classify images into,
    *     only to be specified if includeTop is true.
    * @param includeTop Must be set to true if classifier layers are set.
    */
-  VGGType(const size_t numClasses,
+  VGGType(const size_t numClasses = 1000,
           const bool includeTop = true);
 
   //! Copy the given VGGType.
--- a/models/vgg/vgg.hpp
+++ b/models/vgg/vgg.hpp
@@ -1,0 +1,183 @@
+/**
+ * @file vgg.hpp
+ * @author Shubham Agrawal
+ * 
+ * Definition of VGG models.
+ * 
+ * For more information, kindly refer to the following paper.
+ * 
+ * Paper for VGG.
+ *
+ * @code
+ * @article{Simonyan2014,
+ *  author = {Karen Simonyan, Andrew Zisserman},
+ *  title = {Very Deep Convolutional Networks for Large-Scale Image Recognition},
+ *  year = {2014},
+ *  url = {https://arxiv.org/pdf/1409.1556.pdf}
+ * }
+ * @endcode
+ *
+ * mlpack is free software; you may redistribute it and/or modify it under the
+ * terms of the 3-clause BSD license.  You should have received a copy of the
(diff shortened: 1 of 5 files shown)


models/vgg/vgg.hpp
```diff
@@ -0,0 +1,181 @@
...
+#include <mlpack/methods/ann/init_rules/random_init.hpp>
+#include <mlpack/methods/ann/loss_functions/binary_cross_entropy_loss.hpp>
+
+namespace mlpack {
+namespace models {
+
+/**
+ * Definition of a VGG CNN.
+ * 
+ * NOTE: Note that output size will be 1x1xN. Here, N is number of classes.
```
rcurtin, line 41: Do you mean that we will return a `Cube` for each prediction?  It seems like we should be able to return a length-`N` vector for each input point.

models/vgg/vgg.hpp
```diff
@@ -0,0 +1,181 @@
...
+ * 
+ * @tparam MatType Matrix representation to accept as input and use for
+ *    computation.
+ * @tparam VGGVersion Version of VGG.
+ * @tparam IsBatchNorm Whether to apply Batch Norm layer.
```
rcurtin, line 46: I might suggest either `UsesBatchNorm` or `IsBatchNormalized` (since `BatchNorm` is the name of a layer).  That's a small comment though.

models/vgg/vgg.hpp
```diff
@@ -0,0 +1,181 @@
...
+  /**
+   * VGGType constructor intializes number of classes and weights.
+   *
+   * @param numClasses Optional number of classes to classify images into,
+   *     only to be specified if includeTop is  true.
+   * @param includeTop Must be set to true if weights are set.
+   */
+  VGGType(
+    const size_t numClasses,
+    const bool includeTop = true);
```
rcurtin, line 68:
```suggestion
  VGGType(const size_t numClasses,
          const bool includeTop = true);
```

models/vgg/vgg.hpp
```diff
@@ -0,0 +1,181 @@
...
+{
+ public:
+  //! Create the VGGType layer.
+  VGGType();
+
+  /**
+   * VGGType constructor intializes number of classes and weights.
+   *
+   * @param numClasses Optional number of classes to classify images into,
+   *     only to be specified if includeTop is  true.
```
rcurtin, line 63: I don't think that I understand what `includeTop` does; can you clarify that in the documentation?

models/vgg/vgg.hpp
```diff
@@ -0,0 +1,181 @@
...
+
+  //! Virtual destructor: delete all held layers.
+  virtual ~VGGType()
+  { /* Nothing to do here. */ }
+
+  //! Create a copy of the VGGType (this is safe for polymorphic use).
+  VGGType* Clone() const { return new VGGType(*this); }
+
+  /**
+   * Get Layers of the model.
```
rcurtin, line 87: I would write `Get the FFN object representing the network` just for clarity.

models/vgg/vgg.hpp
```diff
@@ -0,0 +1,181 @@
...
+    vgg->Add(this);
+    return vgg;
+  }
+
+  //! Serialize the VGGType.
+  template<typename Archive>
+  void serialize(Archive& ar, const uint32_t /* version */);
+
+ private:
+  void makeModel()
```
rcurtin, line 109:
```suggestion
  void MakeModel()
```
Just a stylistic convention. :)  Also I would move the implementation into `vgg_impl.hpp` to keep the header file shorter.

models/vgg/vgg.hpp
```diff
@@ -0,0 +1,181 @@
...
+  {
+    std::map<size_t, std::vector<size_t>> const construct {
+      { 11, {64, 0, 128, 0, 256, 256, 0, 512, 512, 0, 512, 512, 0} },
+      { 13, {64, 64, 0, 128, 128, 0, 256, 256, 0, 512, 512, 0, 512, 512, 0} },
+      { 16, {64, 64, 0, 128, 128, 0, 256, 256, 256, 0, 512, 512, 512, 0, 512,
+          512, 512, 0} },
+      { 19, {64, 64, 0, 128, 128, 0, 256, 256, 256, 256, 0, 512, 512, 512, 512,
+          0, 512, 512, 512, 512, 0} }
```
rcurtin, line 117: Can you add a comment indicating what this represents?  I believe it is the network structure of each version (A, A-LRN, B, C, D, E) but I am not quite sure what the `std::vector<size_t>` associated with each model is.

rcurtin: I guess that `0` indicates a `MaxPooling` layer.  Anyway, it would be helpful to describe that in a comment.

models/vgg/vgg.hpp
```diff
@@ -0,0 +1,181 @@
...
+namespace models {
+
+/**
+ * Definition of a VGG CNN.
+ * 
+ * NOTE: Note that output size will be 1x1xN. Here, N is number of classes.
+ * 
+ * @tparam MatType Matrix representation to accept as input and use for
+ *    computation.
+ * @tparam VGGVersion Version of VGG.
```
rcurtin, line 45: It would probably be a good idea to write which versions are valid here.  I guess it must be `11`, `13`, `16`, and `19`; but how would the user get `A-LRN` or `C`?  Is that done by setting the batch normalization parameter to true?

models/vgg/vgg.hpp
```diff
@@ -0,0 +1,181 @@
...
+    typename OutputLayerType = ann::CrossEntropyError,
+    typename InitializationRuleType = ann::RandomInitialization
+  >
+  ann::FFN<OutputLayerType, InitializationRuleType, MatType>* GetModel()
+  {
+    ann::FFN<OutputLayerType, InitializationRuleType, MatType>* vgg =
+        new ann::FFN<OutputLayerType, InitializationRuleType, MatType>();
+    vgg->Add(this);
+    return vgg;
+  }
```
rcurtin, line 102: Is the user supposed to call `GetModel()` and then use `Train()` and `Predict()`?  Or should we provide a `Train()` and `Predict()` function here that simply call the internal model?

(I am not 100% familiar with the models repository, so maybe all the models are like this, I am not sure.)

rcurtin (commented):
Nice! The general implementation looks correct to me.  However I suspect that this is really, really painful to run with OpenBLAS instead of on a GPU. :)  Have you managed to run it at all?

It would be nice to add some simple tests too, so that we know that VGG compiles and runs.  Of course these would have to be very simple tests (like making sure that we can do a forward pass on one point, and a backwards pass on one point, and that the output is a reasonable size), so that the test does not take five years to run. :)

models/vgg/vgg.hpp
shubham1206agra: I left that comment by mistake.

models/vgg/vgg.hpp
shubham1206agra: Idts I have implemented `A-LRN`.

models/vgg/vgg.hpp
shubham1206agra: No need as it is contained inside FFN generated by `GetModel()`.

models/vgg/vgg.hpp
```diff
@@ -0,0 +1,143 @@
...
+class VGGType : public ann::MultiLayer<MatType>
+{
+ public:
+  //! Create the VGGType layer.
+  VGGType();
+
+  /**
+   * VGGType constructor intializes number of classes and weights.
+   *
+   * @param numClasses Optional number of classes to classify images into,
```
rcurtin, line 60: There's no default, so is this optional?  Maybe it makes sense to have `includeTop` before `numClasses`.

models/vgg/vgg.hpp
rcurtin: Understood, thanks!

models/vgg/vgg.hpp
```diff
@@ -0,0 +1,143 @@
...
+#include <mlpack/methods/ann/ffn.hpp>
+#include <mlpack/methods/ann/layer/layer_types.hpp>
+#include <mlpack/methods/ann/init_rules/random_init.hpp>
+#include <mlpack/methods/ann/loss_functions/binary_cross_entropy_loss.hpp>
+
+namespace mlpack {
+namespace models {
+
+/**
+ * Definition of a VGG CNN.
```
rcurtin, line 39: Actually it's probably better to put the paper reference here with the class, since users probably will look for documentation for the class and not necessarily the header file.

Do you think you can add a note on how `VGGVersion` relates to Table 1 in the paper?  i.e. `11` is equivalent to the `A` network, etc., etc.

rcurtin (commented):
Everything is looking good to me here, I only have a few tiny change requests, and then I think it's ready to merge if you're done with it. :+1:

rcurtin:
Ah, oops, I forgot, we also talked about adding some simple tests:

> It would be nice to add some simple tests too, so that we know that VGG compiles and runs. Of course these would have to be very simple tests (like making sure that we can do a forward pass on one point, and a backwards pass on one point, and that the output is a reasonable size), so that the test does not take five years to run. :)

models/vgg/vgg.hpp
shubham1206agra: Writing error :)

models/vgg/vgg.hpp
shubham1206agra: Idts they will look for paper here.

rcurtin: It depends on the tool they are using; if they are just looking through files, I agree, but if they are using some IDE that points them to the class documentation, they will end up looking at this particular comment.  It is a minor issue, so if you feel strongly about it, it's okay to leave it how it is, but in my opinion it would be better placed here. :+1:

tests/vgg_tests.cpp
```diff
@@ -0,0 +1,507 @@
...
+template<typename LayerType>
+void ModelSerializationTest(LayerType& layer)
+{
+  arma::mat input(224 * 224 * 3, 10, arma::fill::randu);
+  arma::mat output(1000, 10, arma::fill::randu);
+
+  ann::FFN<> model;
+  model.Add<LayerType>(layer);
+
+	model.InputDimensions() = std::vector<size_t>({224, 224, 3});
```
rcurtin, line 84: Looks like there are some style issues, with tab characters appearing here and in the later tests.  Can you fix those?

tests/vgg_tests.cpp
```diff
@@ -0,0 +1,507 @@
...
+
+  ens::StandardSGD opt(0.1, 1, 5, -100, false);
```
rcurtin, line 86: Might be nice to add a comment that this only takes one pass over the data (so it does not take super long).

rcurtin (approved):
Nice, thanks for adding the tests.  I don't think I have any more comments; if you can fix the style issues, I'll go ahead and merge. :+1:

rcurtin:
Awesome work @shubham1206agra! :+1: Just two more to merge :)

PR #81, opened 2022-10-05
Refactor models repository for mlpack 4 changes

rcurtin:
Mostly this removes the explicit uses of namespaces and changes headers to include only `mlpack.hpp`.

The tests do not all currently work---but that's not a change from the current state of the repository.  Some of the models still need to be adapted to the ann-vtable transition from a handful of months back.

I also had to make a couple small changes to get some of the tests (which are currently commented out) to compile at all.

In the future I hope to uncomment these tests one by one and get them working again too.


rcurtin:
Okay, I got it to work!  It took some amount of CMake and build system changes to re-enable the OS X and Windows builds too.

zoq:
Any update on the OVH binder issue?

rcurtin:
Not going well, still waiting on a response.  I sent another ping.  If I hear nothing I'll open an issue on a repository somewhere.

PR #85, opened 2024-02-26
Update MobileNet Model

SIDHARTH-7:
This PR is to update the Implementation of MobilenetV1 which was implemented in #72 and worked upon by @Aakash-kaushik , @zoq and @kartikdutt18 . The files are named just mobilenet so that we could include further versions in the same files for better readability.


SIDHARTH-7:
@shrit I will add some tests and then we could check the Implementation of Separable Convolution as well on my local machine.
Please check if my current Implementation can be Improved or if it is wrong.

.vscode/settings.json
```diff
@@ -0,0 +1,82 @@
+{
```
rcurtin, line 1: No need to add this file---we could add it to the `.gitignore` perhaps?

models/mobilenet/mobilenet.hpp
```diff
@@ -0,0 +1,142 @@
...
+/**
+ * Definition of a MobileNet CNN.
+ * 
+ * @tparam OutputLayerType The output layer type used to evaluate the network.
+ * @tparam InitializationRuleType Rule used to initialize the weight matrix.
+ */
+template<
+  typename MatType = arma::mat
+>
+class MobileNetType : public MultiLayer<MatType>
```
rcurtin, line 47: I have a general question about the design.  I think it might be cleaner to simply have `MobileNetType` internally store an `FFN` that you add the layers to in the constructor.  I think several of the other models are built in this way.  Is there any particular advantage of this approach that I overlooked?  It seems like if you did just hold an `FFN` internally, you could use default copy/move constructors and operators and reduce the amount of code that you need.

models/mobilenet/mobilenet.hpp
```diff
@@ -0,0 +1,142 @@
...
+  std::map<size_t, size_t> mobileNetConfig = {
+                                                {128, 2},
+                                                {256, 2},
+                                                {512, 6},
+                                                {1024, 2},
+                                              };
+}; // MobileNetType class
+
+// convenience typedef.
+typedef MobileNetType<arma::mat> Mobilenet;
```
rcurtin, line 133:
```suggestion
typedef MobileNetType<arma::mat> MobileNet;
```

tests/mobilenet_tests.cpp
```diff
@@ -0,0 +1,147 @@
...
+TEST_CASE("MobileNetNoTopTest", "[MobileNetTests]")
+{
+  arma::mat input(224 * 224 * 3, 10, arma::fill::randu);
+  arma::mat output;
+  models::Mobilenet mobilenetLayer(512, false);
+  FFN<> model;
+  model.InputDimensions() = std::vector<size_t>({224, 224, 3});
+  model.Add<models::Mobilenet>(mobilenetLayer);
+  ModelDimTest(model, input, 9216, 10);
+}
```
rcurtin, line 147: It would be really great to add a test for `models::MobileNetType<arma::fmat>`, just to make sure that the model works with low precision.  I think so long as you don't test serialization, that should work just fine.

(The reason serialization would have a problem is because you'd need to register all the `arma::fmat` layers for serialization.  That can add a lot of compile time... but I think it's safe to say that if MobileNet serializes correctly with `arma::mat`, then it will serialize correctly with `arma::fmat`, assuming that each layer correctly serializes with `arma::fmat`.  From the perspective of this code, that should be a safe assumption---testing individual layer serialization should be done inside of mlpack's core code.)

rcurtin (commented):
Hey there @SIDHARTH-7, really nice to have this support added.  Sorry that this took almost four months to get to.  I just have some high-level comments for now; I have not checked the implementation in detail.  I know my review is late, but if you are still interested in working on it so much later, let me know what you think. :+1:

